PR #1, opened 2019-07-28
Add authentication instructions to README

UnsolvedCypher:
Just adding what Jonatan said in Slack to the README so the info is easier to find in the future :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,3 +25,7 @@ cd client
 npm install
 npm start
 ```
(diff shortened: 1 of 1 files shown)


jonatanklosko (approved):
As said I plan to figure out how to make OAuth sign in more straightforward in development and provide better description. Thanks adding instructions for current setup! :rocket:

PR #51, opened 2020-02-07
Cumulative time limit (issue #5)

danieljames-dj:
Just like converting the attempt to DNF, this will convert the last time(s) to DNF if cumulative time limit is crossed. This will allow the delegate to know while data entry that cumulative time limit is exceeded.

### DIFF
--- a/client/src/components/admin/ResultForm/ResultForm.js
+++ b/client/src/components/admin/ResultForm/ResultForm.js
@@ -121,9 +121,29 @@ const ResultForm = ({
             initialValue={attempt}
             disabled={!result || index >= disabledFromIndex}
             onValue={value => {
-              const updatedValue =
-                timeLimit && value >= timeLimit.centiseconds ? -1 : value;
-              const updatedAttempts = setAt(attempts, index, updatedValue);
+              var updatedAttempts = setAt(attempts, index, value);
+              if (timeLimit) {
+                if (timeLimit.cumulativeRoundIds.length === 1) {
+                  var totalTime = 0,
+                    i;
+                  for (i = 0; i < updatedAttempts.length; i++) {
+                    if (updatedAttempts[i] > 0) {
+                      totalTime += updatedAttempts[i];
+                      if (totalTime >= timeLimit.centiseconds) {
+                        updatedAttempts[i] = -1;
+                        break;
+                      }
+                    }
+                  }
+                  for (; i < updatedAttempts.length; i++) {
+                    if (updatedAttempts[i] > 0) {
+                      updatedAttempts[i] = -1;
+                    }
+                  }
+                } else if (value >= timeLimit.centiseconds) {
+                  updatedAttempts[index] = -1;
+                }
+              }
               setAttempts(
                 meetsCutoff(updatedAttempts, cutoff, eventId)
                   ? updatedAttempts
--- a/client/src/components/admin/ResultForm/ResultForm.js
+++ b/client/src/components/admin/ResultForm/ResultForm.js
@@ -16,7 +16,7 @@ import {
   formatAttemptResult,
   attemptsWarning,
   applyTimeLimit,
-  applyCutOff,
+  applyCutoff,
 } from '../../../logic/attempts';
 import { best, average } from '../../../logic/stats';
 import { cutoffToString, timeLimitToString } from '../../../logic/formatters';
@@ -64,7 +64,7 @@ const ResultForm = ({
 
   const submissionWarning = attemptsWarning(attempts, eventId);
 
-  const disabledFromIndex = meetsCutoff(attempts, cutoff, eventId)
+  const disabledFromIndex = meetsCutoff(attempts, cutoff)
     ? solveCount
     : cutoff.numberOfAttempts;
 
@@ -124,9 +124,9 @@ const ResultForm = ({
             disabled={!result || index >= disabledFromIndex}
             onValue={value => {
               setAttempts(
-                applyTimeLimit(
-                  applyCutOff(setAt(attempts, index, value), cutoff, eventId),
-                  timeLimit
+                applyCutoff(
+                  applyTimeLimit(setAt(attempts, index, value), timeLimit),
+                  cutoff
                 )
               );
             }}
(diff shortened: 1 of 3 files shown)


client/src/components/admin/ResultForm/ResultForm.js
```diff
@@ -121,9 +121,29 @@ const ResultForm = ({
...
+                        break;
+                      }
+                    }
+                  }
+                  for (; i < updatedAttempts.length; i++) {
+                    if (updatedAttempts[i] > 0) {
+                      updatedAttempts[i] = -1;
+                    }
+                  }
+                } else if (value >= timeLimit.centiseconds) {
```
jonatanklosko: The first condition is not satisfied for cross-cumulative time limits (i.e. `cumulativeRoundIds.length > 1`), in which case we definitely don't want to apply the time limit to each of the solves, so this should go in pair with `cumulativeRoundIds.length === 0`.

client/src/components/admin/ResultForm/ResultForm.js
```diff
@@ -121,9 +121,29 @@ const ResultForm = ({
             initialValue={attempt}
             disabled={!result || index >= disabledFromIndex}
             onValue={value => {
-              const updatedValue =
-                timeLimit && value >= timeLimit.centiseconds ? -1 : value;
-              const updatedAttempts = setAt(attempts, index, updatedValue);
+              var updatedAttempts = setAt(attempts, index, value);
```
jonatanklosko: Please use `const` everywhere to stay consistent across the project (either way `var` should be avoided since `let` and `const` introduction, as these are block-scoped and thus most predictable).

While on that, functional approach is preferable here (partially enforced by React, but I'd rather stick to it in general), so please avoid any sort of mutations and imperative-style computation ^^

On a side-note, as this is quite some logic now, it would be best to move it to another module (`logic/attempts.js`), here's how I'd go about it (iterating over the attempts, keeping track of the sum, and producing an updated list of attempts):
```js
const applyTimeLimit = (attempts, timeLimit) => {
  if (timeLimit.cumulativeRoundIds.length === 0) {
    return attempts.map(attempt =>
      attempt >= timeLimit.centiseconds ? -1 : attempt
    );
  } else {
    /* Note: for now cross-round cumulative time limits are handled
       as single-round cumulative time limits for each of the rounds. */
    const [updatedAttempts, sum] = attempts.reduce(
      ([updatedAttempts, sum], attempt) => {
        const updatedSum = attempt > 0 ? sum + attempt : sum;
        const updatedAttempt =
          attempt > 0 && updatedSum >= timeLimit.centiseconds ? -1 : attempt;
        return [updatedAttempts.concat(updatedAttempt), updatedSum];
      },
      [[], 0]
    );
    return updatedAttempts;
  }
};
```
(for cross-cumulative time limits it's still better to check them within each round, than do nothing)

Then here it would become
```js
const updatedAttempts = applyTimeLimit(
  setAt(attempts, index, value),
  timeLimit
);
```
As you can see all the functions return new arrays, keeping the arguments untouched.

If you feel like it, you can also define a similar `applyCutoff` function to move the cutoff logic outside =)

Also please write some tests for that in `logic/tests/attempts.test.js` (`meetsCutoff` may be a good example to base on).

jonatanklosko (commented):
Hey Daniel, thanks for contributing and proposing the solution! :rocket: Leaving some comments for you, let me know if you have any doubts about them!

client/src/components/admin/ResultForm/ResultForm.js
danieljames-dj: Hi Jonathan, thanks a lot for the detailed comments. It was really helpful.
I'm a bit confused because I couldn't find the code you wrote in the comment in the latest code. May I know whether you wrote it for me so that I can use that function?
If so, I'll write it and write a similar apply cutoff too as you said. Can I proceed?

jonatanklosko: Hey, none of that code already exists! All I meant was to define the `applyTimeLimit` function in the `attemtps.js` module and then use it [here](https://github.com/thewca/wca-live/blob/d7666dd3fb69426587b207a8ba8aec474e9df077/client/src/components/admin/ResultForm/ResultForm.js#L124-L126). I suggested defining similar `applyCutoff` to move [this](https://github.com/thewca/wca-live/blob/d7666dd3fb69426587b207a8ba8aec474e9df077/client/src/components/admin/ResultForm/ResultForm.js#L128-L132) logic into the `attempts.js` module as well. [These](https://github.com/thewca/wca-live/blob/d7666dd3fb69426587b207a8ba8aec474e9df077/client/src/logic/tests/attempts.test.js#L143-L166) are the tests that may be a good start for testing the new functions =)

danieljames-dj:
Hi @jonatanklosko , I've made the changes as you suggested. Thanks a lot for your suggestions. It was a great lesson for me in good practices of coding. I would like to take up more challenges.
Few comments:
1. I've just copy-pasted your applyTimeLimit code because I couldn't find any improvements.
2. Earlier, while calling `meetsCutoff` in ResultForm, one extra parameter eventId was passed, but it was not used in the `meetsCutoff` of attempts.js. I didn't remove it as I was not sure it can be removed. Please let me know if I can remove the extra parameter from all the places the function is used (it is used in `disabledFromIndex` as well).
3. After writing the test cases, I tried running the test using `jest attempts.test.js`, but I got the error `SyntaxError: Cannot use import statement outside a module`. So, I couldn't test the test cases I wrote.

jonatanklosko (commented):
Thanks for writing the tests!

Sure, please remove the unnecessary argument, it must be some leftover ^^

As for running the tests, did you try `npm run test`? That should execute the tests in convenient watch mode (by default it runs tests for files that changed, so you can press `a` to run all the tests). Let me know if for some reason it doesn't work for you!

danieljames-dj:
Hi @jonatanklosko I've tried my best to implement all the fixes that you suggested. I hope I didn't miss anything. Please let me know if I missed anything.
Also, test is working with `npm run test`, thanks a lot for that too. :-)

jonatanklosko (approved):
Perfect, thanks Daniel! :ship: :tada:

PR #52, opened 2020-02-10
Add unfinished events to podium list

### LINKED ISSUE #46: Show empty podium for events instead of not showing event at all
An issue we ran into at Weston-super-Mare Open 2019 was quite funny. For Skewb and Square-1 we had both events as a Combined Final. However WCA Live auto-deletes competitors from previous rounds upon opening a subsequent round. Because of this, we did not manually remove the no-shows for these events. When it came to the award ceremony, it was exclaimed that we had missed out Square-1 and Skewb and we realised the reason immediately. 

I think once the first event has been completed, it would be good on the podium page to just show something along the lines of "**Undecided: Final round still in progress!**" under the events that don't have finalised rounds to avoid missing events.

saranshgrover:
Fixes #46

### DIFF
--- a/client/src/components/Podiums/Podiums.js
+++ b/client/src/components/Podiums/Podiums.js
@@ -63,7 +63,6 @@ const Podiums = ({ match }) => {
   if (loading && !data) return <Loading />;
   if (error) return <ErrorSnackbar />;
   const { competition } = data;
-  console.log(competition);
 
   return (
     <Fragment>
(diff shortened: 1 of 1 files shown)


jonatanklosko:
Could you please include a screenshot for the reference?

client/src/components/Podiums/Podiums.js
```diff
@@ -56,6 +63,7 @@ const Podiums = ({ match }) => {
   if (loading && !data) return <Loading />;
   if (error) return <ErrorSnackbar />;
   const { competition } = data;
+  console.log(competition);
```
jonatanklosko: Nit: this shouldn't be here =)

saranshgrover:
image

client/src/components/Podiums/Podiums.js
```diff
@@ -75,6 +83,29 @@ const Podiums = ({ match }) => {
               />
             </Grid>
           ))}
+          {competition.events.length !== competition.podiums.length && (
+            <Grid item key={`events_not_completed`}>
```
jonatanklosko: The `key` prop is necessary only when rendering a list (usually with `Array#map`) and that's not that we do here =)

client/src/components/Podiums/Podiums.js
```diff
@@ -75,6 +83,29 @@ const Podiums = ({ match }) => {
...
+              <Typography>{`Podiums for the following events are yet to be determined:`}</Typography>
+              <Grid container direction="row" spacing={3}>
+                {competition.events
+                  .filter(
+                    event =>
+                      competition.podiums.filter(
+                        podium => podium.event.id === event.id
+                      ).length === 0
```
jonatanklosko: As we test for the presence only, `.filter(...).length === 0` could be replaced with `.some(...)` :cat:

jonatanklosko (commented):
Thanks for contributing with the clean implementation, just few comments =)

jonatanklosko (approved):
LGTM, thanks! :tada:

PR #91, opened 2022-02-13
Adhere to WCIF changes

### LINKED ISSUE #74: Allow Trainee Delegates to Manage Competitions
Trainee Delegates are not currently automatically added as admins to WCA Live competition pages. Ideally they should be automatically listed in this role so that they are able to perform tasks such as data checking and entering without the need to add them as a specific score taker for the competition.

jonatanklosko:
See https://github.com/thewca/wcif/pull/3 and https://github.com/thewca/wcif/pull/4.

Once https://github.com/thewca/worldcubeassociation.org/pull/6690 is merged, this closes #74.

### DIFF
--- a/server/lib/wca_live/competitions/qualification.ex
+++ b/server/lib/wca_live/competitions/qualification.ex
@@ -6,20 +6,20 @@ defmodule WcaLive.Competitions.Qualification do
   use WcaLive.Schema
   import Ecto.Changeset
 
-  @required_fields [:type, :when, :attempt_result]
+  @required_fields [:type, :when_date, :level]
   @optional_fields []
 
   @primary_key false
   embedded_schema do
     field :type, :string
-    field :when, :utc_datetime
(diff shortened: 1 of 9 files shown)


server/lib/wca_live/competitions/qualification.ex
```diff
@@ -6,20 +6,20 @@ defmodule WcaLive.Competitions.Qualification do
...
   @optional_fields []
 
   @primary_key false
   embedded_schema do
     field :type, :string
-    field :when, :utc_datetime
-    field :attempt_result, :integer
+    field :when_date, :date
+    field :level, :integer
   end
```
jonatanklosko, line 17: Note to self: the qualification is stored as a JSON column in the database, and currently it's empty for all competitions (since the WCA website doesn't support it yet), therefore it's safe to change the schema without any migrations.

PR #111, opened 2022-09-26
Advancing when round is in progress: don't take the no-shows into account. Fixes #109

### LINKED ISSUE #109: Add warning "Results are incomplete and may change"
Hello software team,

For a competitor it can be very frustrating to find out they did not advance to the next round even though they had previously checked the live results and thought they qualified. This can happen for example when some results are entered late or when the number of advancing competitors is changed due to no-shows being removed (if the advancement criterion is set to a percentage of total competitors). Unfortunately the Disciplinary Committee is relatively often notified about this, either because a parent complains directly to us about the situation or because a delegate reports a heated discussion during a competition.

I think we could help mitigate this by adding a simple warning when the results are not completely entered. I could see this consist of two/three parts:

- On the result page for a round, right below the name of the event, a warning, possibly highlighted in yellow or red: "Results for this round are incomplete. Rankings and qualification limits may change."
- On the competitor's result page, a question mark next to rankings for incomplete rounds, showing the same message as above when clicked or tapped.
- Same as the previous point on the mobile version when clicking/tapping on a competitor's name from the round results page.

Another idea is to not highlight any ranking in green before the results are complete. This is compatible, but not dependent on, the warning message proposal I explained above.

I hope you can consider implementing something like this soon, but I understand that you are probably busy with more important stuff. If you can think of a different way to solve the same problem feel free to implement that instead :)

Thank you,
Sebastiano

Goosly:
Dear WST, @gregorbg , @jonatanklosko ,

The scenario is the following:
100 competitors registered for 3x3 round 1. 75% advance to the next round. My scoretakers enter results, and WCA Live shows 75 competitors advance to the next round, with the green label on their ranking. However, 5 people did not show up. My scoretakers remove those 5 people. Now, only 71 competitors advance to the next round. The competitors ranked 72nd, 73th, 74th and 75th are terribly disappointed because WCA Live tricked them into believing they were going to round 2.

This has been a returning issue at hundreds of competitions worldwide and yesterday, we had the same issue with one person thinking he was going to round 2 and then we had to remove him. One of the organizers then decided to give up his spot in round 2, so the other competitor could be added to round 2 instead.

This PR attempts to fix that. However, I don't know Euphoria and I did not setup anything locally to run the tests :-) So I hope you can help me if the tests are not working, or if I did anything else wrong. Thanks in advance for your time!

### DIFF
--- a/server/lib/wca_live/scoretaking/advancing.ex
+++ b/server/lib/wca_live/scoretaking/advancing.ex
@@ -67,11 +67,16 @@ defmodule WcaLive.Scoretaking.Advancing do
         %{results: results, advancement_condition: advancement_condition} = round
         format = Format.get_by_id!(round.format_id)
 
+        # We ignore unranked results until they are entered. For percentage-based
+        # advancement rules this means that we qualify less results initially and
+        # start qualifying more as the missing results are entered. This way, in
+        # case the remaining missing results are quit, we don't un-qualify anyone
+        results = Enum.filter(results, & &1.ranking)
+
         # See: https://www.worldcubeassociation.org/regulations/#9p1
         max_qualifying = floor(length(results) * 0.75)
-        rankings = Enum.filter(results, fn result ->
-          result.best != nil
-        end) |> Enum.map(& &1.ranking) |> Enum.reject(&is_nil/1) |> Enum.sort()
+
+        rankings = results |> Enum.map(& &1.ranking) |> Enum.sort()
 
         first_non_qualifying_ranking =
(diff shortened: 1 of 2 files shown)


jonatanklosko (approved):
Hey @Goosly! Thanks for following up on the issue, I just pushed some adjustments and this is good to go :)

Goosly:
I've tested this on an upcoming competition with mock results (then cleared the round afterwards) and it works as expected.

Thank you for the extremely quick help @jonatanklosko ! ❤️

PR #112, opened 2022-09-27
Add 'level' property back to Qualification encoding

gregorbg:
Was accidentally deleted in https://github.com/thewca/wca-live/pull/104

In particular, synchronizing competitions that had qualification results caused `Validation failed: Qualification is invalid` errors to pop up, because the JSON was serialized with `"level": null`.

I verified locally that this fix works by testing with `SpanishChampionship2022`. Importing, then immediately re-syncing without any changes yields the error without my fix, but works with the fix from this patch. I didn't bother setting up the test environment as this is my first PR and it's such a trivial fix, but I really wonder why the automated test suite didn't catch this?!

### DIFF
--- a/server/lib/wca_live/competitions/qualification.ex
+++ b/server/lib/wca_live/competitions/qualification.ex
@@ -6,7 +6,7 @@ defmodule WcaLive.Competitions.Qualification do
   use WcaLive.Schema
   import Ecto.Changeset
 
-  @required_fields [:type, :result_type, :when_date]
+  @required_fields [:type, :result_type, :when_date, :level]
   @optional_fields []
 
   @primary_key false


jonatanklosko (approved):
Great catch, thank you!

> I really wonder why the automated test suite didn't catch this

The validation error comes from the WCA website. Without this change the level field would be ignored when importing the WCIF, and then we would submit the new WCIF with qualification that fails a validation there.

gregorbg:
Thanks for merging straight away!

> The validation error comes from the WCA website. Without this change the level field would be ignored when importing the WCIF, and then we would submit the new WCIF with qualification that fails a validation there.

I get that the validation comes from the WCA website. I guess the answer to my question is that it wasn't caught by the test suite because there is no test (correct me if I'm wrong) that de-codes and re-encodes a WCIF like it happens when we sync.

jonatanklosko:
@gregorbg ah I see what you mean. We have separate tests for import and export, however they don't assert on every single field.

https://github.com/thewca/wca-live/blob/19476f01f4b138e0dca72cd66ab3e96e3e3c8a9b/server/test/wca_live/synchronization/import_test.exs#L184

Here you can see we verify that the qualification gets imported, but we don't check `level` specifically.

I added an additional integration test in #113 to make sure import and export WCIFs match as expected, good call :)

PR #113, opened 2022-09-27
Add integration test verifying WCIF consistency in import/export

jonatanklosko:
To avoid inconsistencies as in #112.


gregorbg:
Thanks for adding this additional integration test!

Why doesn't this fail with the recent WCIF spec update to include Series information? As far as I'm concerned, WCA Live has no model for Series, so they should not be part of the re-encoded WCIF and the test should fail. But the CI checks on this PR pass. What am I overlooking?

jonatanklosko:
@gregorbg the test uses a hard-coded WCIF and it didn't include the series information. Currently we don't import it either way, I've just added an explicit note about it. Does that make sense?

gregorbg (approved):

gregorbg:
Alright, thanks for the explanation! :smile:

Issue #121, opened 2022-11-15
Change "PB" to "PR"  in personal results section


dunkOnIT:
As this [forum thread](https://forum.worldcubeassociation.org/t/changing-wca-live-to-pr-instead-of-pb/28111) notes, the emerging convention is to use "PB" (personal best) for one's best time overall, and "PR" for best time achieved in competition.

While I don't think the current wording is "highly misleading", I do think PB/PR is a helpful convention that I'm assuming it would be trivial to implement.

kr-matthews:
WCA profiles literally say "Current Personal Records" so PR is technically correct; the current usage of PB in WCA Live is inconsistent with the WCA site. PR is also consistent with WR, CR, and NR.

![image](https://user-images.githubusercontent.com/49137025/201976023-fcd94379-1ebd-4f80-9969-53cd50e6e044.png)

jonatanklosko:
Closed in #122.

PR #122, opened 2022-11-19
Replace PB with PR

nsilvestri:
This PR changes the RecordTag to read PR instead of PB. For the sake of internal consistency, it also replaces references of "pb" and its variations throughout the code. Addresses #121.

Preview: 
image

### DIFF
--- a/client/src/components/RecordTag/RecordTag.js
+++ b/client/src/components/RecordTag/RecordTag.js
@@ -15,7 +15,7 @@ const styles = {
     color: (theme) => theme.palette.getContrastText(green['A400']),
     backgroundColor: green['A400'],
   },
-  pb: {
+  pr: {
     color: (theme) => theme.palette.getContrastText(blue[700]),
     backgroundColor: blue[700],
   },
--- a/client/src/components/RecordTagBadge/RecordTagBadge.js
+++ b/client/src/components/RecordTagBadge/RecordTagBadge.js
@@ -2,8 +2,8 @@ import React from 'react';
 import { Box } from '@mui/material';
 import RecordTag from '../RecordTag/RecordTag';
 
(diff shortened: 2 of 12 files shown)


jonatanklosko (approved):
@nsilvestri awesome, thanks!

Randomno:
Because of caching issues (I think), PRs appeared on event pages like this:
image

The feedback to this was quite positive so it's probably worth actually implementing.

jonatanklosko:
@Randomno it was like this for a brief moment, because I realised we need to update the database as well, but it should be all good now :)

nsilvestri:
@jonatanklosko just to clarify, I think they were pointing out the presence of the PR tag on the round results table, not the issues with the PR tag styles getting cached. I'm not sure what database changes caused it, but this accidental "feature" was received very positively on Cyoubx's Friends while it was live.

It should be fairly simple to implement. I think removing the `hidePr` prop on [RoundResultsTable.js](https://github.com/thewca/wca-live/blob/1dd3cb8678fb284330075f2d9bcc2864346166bc/client/src/components/RoundResults/RoundResultsTable.js#L132) should accomplish this, if I understand the code correctly. I can put together a pull request in a few hours if this is a desired change.

jonatanklosko:
@nsilvestri ohh I see, sounds good to me then! We also need to remove `hidePr` from the corresponding [dialog](https://github.com/thewca/wca-live/blob/1dd3cb8678fb284330075f2d9bcc2864346166bc/client/src/components/RoundResults/RoundResultDialog.js#L70) we show on mobile, a PR is very welcome :)

Sidenote: looking again at the screenshot it may've actually been a caching issue, because the value "PR" comes from the database, but the client doesn't recognise it. We used to have a service worker not so long ago and it would effectively defer client updates; it's been removed, but I imagine for some people it may not have uninstalled yet.

PR #124, opened 2022-11-21
Show PR badge on round results table and dialog

nsilvestri:
The PR badges were accidentally shown on the results table previously, but it generated a good response in the short time that the "feature" was live. There was some discussion on this at [this pull request](https://github.com/thewca/wca-live/pull/122). This commit officially changes the round results table and the round result dialog to show the PR badges when relevant. It does not change the admin results table (displayed while inserting times) nor does it change the results projector view, both of which continue to hide the PR badge.

### Round Results Table

image

### Round Result Dialog

image


jonatanklosko (approved):
Thanks!

Issue #126, opened 2022-11-26
Change design of "PR" tag to be more easily distinguishable from NR/CR


dunkOnIT:
Again referencing [this forum thread](https://forum.worldcubeassociation.org/t/changing-wca-live-to-pr-instead-of-pb/28111/11), the change from "PB" to "PR" has been implemented, but as multiple users note, it looks a bit confusing as the design is very similar to the NR/CR tag. 

@nsilvestri would you be interested in an update to #122 where you make a minor design change to the tag? My first thought would be to make it more opaque and thus less attention-grabbing than the NR/CR tags (I don't think a pure colour change would be enough, as I think NR/CR are already different colours?). I'm not a UI designer, though, so up to you :D

nsilvestri:
Thanks for the feedback Duncan, I've been monitoring the response to this as well. I've thrown together a quick alternative, and I'd appreciate feedback on it. Note that these styles apply globally to the PR tag, not just on the round results table. 

Dark mode:

![image](https://user-images.githubusercontent.com/33607815/204075798-a0fc33b8-b3e0-4f3f-a829-59b21893f0eb.png)


Light mode:

![image](https://user-images.githubusercontent.com/33607815/204075810-ee4144ce-b2cc-4bb5-b64a-0b308d67bc33.png)

jonatanklosko:
@nsilvestri I think there isn't enough contrast between the tag background and page background. I would try reversing the background and font colour, so on light mode it can be a dark grey background and white font.

nsilvestri:
@jonatanklosko here's the inverse of the previous iteration:

![image](https://user-images.githubusercontent.com/33607815/204098436-3ce82e70-3daf-4a11-abee-a3aa0ae0775f.png)

![image](https://user-images.githubusercontent.com/33607815/204098455-82a7954b-2396-472a-b2a3-242798812b31.png)

In my own opinion, this is about equally as attention-grabbing as the original iteration, especially on light mode. Maybe the solution to this problem is keeping the blue PR tags, but hiding them behind a user-controlled switch that toggles showing/hiding PRs? The switch would pass a `hidePr` prop to the `RecordTagBadge`s on the `RoundResultsTable`.

jonatanklosko:
Yeah, I agree that feel to much, but if there are are a lot of PRs I think any colour will have this issue. So another alternative I can think of is removing the background entirely, with a little less padding; this way it wouldn't stand out, but the information would be there.

[...]

PR #127, opened 2022-11-28
Add litePr flag and use on RoundResultsTable

nsilvestri:
When many PRs were broken, the blue PR badges would cover the round results table. This commit adds an alternative style called `litePr` which has a background that matches the current color scheme: light on light mode and dark on dark mode. This new style is used on the round results table, so that the PRs still appear on the round results table in a less distracting appearance. Otherwise, the standard blue badges are used for PR in all other contexts. Addressing #126.

There's a little bit of repeated code in the new styling but making it not repeat would be about the same amount of code. There's also a little bit of prop drilling from `RoundResultsTable` through `RecordTagBadge` to the final `RecordTag`. Please advise if there's a better approach (I have literally never had my React code reviewed by someone else before) :)

Round Results Table:

image
image

Competitor Result Dialog (unchanged):

image

Competitor Results (unchanged):

image

### DIFF
--- a/client/src/components/RecordTag/RecordTag.js
+++ b/client/src/components/RecordTag/RecordTag.js
@@ -1,6 +1,6 @@
 import React from 'react';
 import { Box } from '@mui/material';
-import { red, yellow, green, blue } from '@mui/material/colors';
+import { red, yellow, green, blue, grey } from '@mui/material/colors';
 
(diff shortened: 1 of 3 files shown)


jonatanklosko (approved):
Looks great, thanks @nsilvestri!

PR #140, opened 2023-02-28
Add infrastructure files

jonatanklosko:
More details in `infra/README.md`.

### DIFF
--- a/infra/lb.tf
+++ b/infra/lb.tf
@@ -1,0 +1,109 @@
+resource "aws_security_group" "lb" {
+  name        = "${var.name_prefix}-load-balancer"
+  description = "Production load balancer"
+  vpc_id      = aws_vpc.this.id
+
+  ingress {
+    from_port   = 80
+    to_port     = 80
+    protocol    = "tcp"
+    cidr_blocks = ["0.0.0.0/0"]
+    description = "Allow HTTP"
+  }
+
+  ingress {
+    from_port        = 80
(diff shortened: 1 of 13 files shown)


infra/lb.tf
```diff
@@ -0,0 +1,115 @@
+resource "aws_security_group" "lb" {
+  name        = "${var.name_prefix}-load-balancer"
+  description = "Production load balancer"
+  vpc_id      = aws_vpc.this.id
+
+  ingress {
```
FinnIckler, line 6: are you forwarding port 80 to 443? Otherwise port 80 will just run into nothingness here

FinnIckler (commented):
Just one thing about the LB, otherwise it looks great!

infra/lb.tf
jonatanklosko: Ah that was for testing. If we don't have LB listener on 80 does the SG rule matter?

FinnIckler: it will, because when we look at the monitoring we will be very confused by all the 502 bad gateway errors that happen when someone tries to connect to port 80 on the site. I would just [redirect](https://aws.amazon.com/premiumsupport/knowledge-center/elb-redirect-http-to-https-using-alb/)

jonatanklosko: Sounds good, thanks!

FinnIckler (approved):
LGTM!

PR #148, opened 2023-06-10
Add bulk quit for no-show competitors

### LINKED ISSUE #96: Idea: Make clearing no-shows simpler
When I am doing data entry once I get to the end of a round I go through all the blank scorecards and remove that person from the round - this is a key check to make sure there isn't a missing scorecard or someone's results were entered using the wrong ID during data entry. 

Probably at small comps where there are 5 or so no-shows it is relatively simple, but at big 200+ competitor events there are typically a lot more and there are a lot of steps (click their name, click "Clear", confirm you want to do that) and the additional challenge of a ~10 second lag per person if you have slow-ish internet and a lot of people in the round that need to be removed. Obviously bulk removing them (just opening the following round) is dangerous in case there are missing scorecards etc, and in Australia at least we generally don't open the following round until the earlier round is all double checked. 

Idea, at some point within the data page as you finish data entry for a round, click a button somewhere to enter "clearing mode". This would take you to a list of all the names without results, and you could checkbox through the list to work out which ones to remove, and then click a "clear" button (probably with a confirmation after that) which then removes all those people from the round in one go.

jonatanklosko:
Closes #96.

Allows for bulk-quitting competitors in first rounds:

image

### DIFF
--- a/test/wca_live/scoretaking_test.exs
+++ b/test/wca_live/scoretaking_test.exs
@@ -503,6 +503,25 @@ defmodule WcaLive.ScoretakingTest do
     assert 4 == Enum.count(results, & &1.advancing)
   end
 
+  test "removes_no_shows_from_round/2 removes no-show result with matching ids from the given round" do
+    round = insert(:round, number: 1)
+    insert_list(4, :result, round: round)
+
+    person1 = insert(:person)
+    insert(:result, person: person1, round: round)
+
+    person2 = insert(:person)
+    insert(:result, person: person2, round: round, attempts: [])
+
+    assert {:ok, updated} =
+             Scoretaking.remove_no_shows_from_round(round, [person1.id, person2.id])
+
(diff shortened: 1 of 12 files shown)


test/wca_live/scoretaking_test.exs
```diff
@@ -476,7 +476,9 @@ defmodule WcaLive.ScoretakingTest do
     [next_qualifying_person] = setup.get_people_by_rank.(5)
 
     assert {:ok, updated} =
-             Scoretaking.remove_person_from_round(second_person, setup.second_round, replace: true)
+             Scoretaking.remove_person_from_round(second_person, setup.second_round,
+               replace: true
+             )
```
jonatanklosko:
```suggestion
             Scoretaking.remove_person_from_round(second_person, setup.second_round replace: true)
```

PR #151, opened 2023-07-01
Changes to the Recognition of Regions

gregorbg:
As per Board of Directors request via email. The WST is not making any political statements about China or Chinese diplomacy. We are acting on a binding instruction of our supervising Board authority.

Please direct all relevant discussions and opinions towards the public or internal mail threads. Discussion comments under this PR that go beyond the technical implementation will be disregarded.


jonatanklosko (approved):

gregorbg:
As of writing this comment, the "Upcoming" tab on the homepage at https://live.worldcubeassociation.org/ does not show the intended flag from this PR next to "New Taipei Open 2023"

jonatanklosko:
@gregorbg it shows up as expected for me. Can you try closing the WCA Live tabs and opening again or trying in incognito? It's likely a caching issue.

gregorbg:
That was a caching issue indeed :facepalm: 
But I am still surprised to see that it was a caching issue across reboots on my machine, is this something that you are controlling from the application side? Or is this my browser running haywire?

jonatanklosko:
This was the case for a few other people and one theory I have is that on some devices the old service worker hasn't been deactivated yet (and it would do more aggressive caching). We used to have a service worker but it only caused troubles so I disabled it like 10 months ago. The confusing part is that it should definitely be deactivated if the user visited WCA Live several times since that change. Currently I don't have a better idea, we don't do any unusual caching and all assets have hash in their name.

gregorbg:
Hm, in that case I guess we're as fine as we can reasonably be. Thanks!

PR #157, opened 2023-07-09
Add warning when submitting new world record

nsilvestri:
Using the API added in `bf89d0f`, world records are now fetched and used to check results for new world records when they are submitted.

There were some considersations when altering the warning logic:

1. The purpose of the warning is to have the user go back and double-check the results. Therefore, it doesn't ultimately matter exactly which warning gets shown to the user, if more than one warning is applicable to a set of results.
2. All events have world record singles, but `333mbf` is the only one without an average. The current code has a branch for `333mbf` warnings, so before it branches off, I check for a WR single. After the branch is taken that no longer includes `333mbf`, I check for a new WR average.
3. Only final results from the WCA database are returned from the query. If a world record is broken multiple times before results are submitted, there will be unnecessary warning dialogs. However, since the warning is simply to detect false positives, it's okay to show the dialog eagerly.
4. This does not check for Continental Records or National Records, as that involves more complex logic and data to check the competitor's country and continent.

### DIFF
--- a/client/src/components/admin/AdminRound/AdminRoundContent.js
+++ b/client/src/components/admin/AdminRound/AdminRoundContent.js
@@ -36,7 +36,7 @@ const ENTER_RESULTS = gql`
   ${ADMIN_ROUND_RESULT_FRAGMENT}
 `;
 
-function AdminRoundContent({ round, competitionId }) {
+function AdminRoundContent({ round, competitionId, officialWorldRecords }) {
   const confirm = useConfirm();
   const { enqueueSnackbar, closeSnackbar } = useSnackbar();
   const apolloErrorHandler = useApolloErrorHandler();
@@ -135,6 +135,7 @@ function AdminRoundContent({ round, competitionId }) {
             focusOnResultChange={true}
             disabled={loading}
             onSubmit={handleResultAttemptsSubmit}
+            officialWorldRecords={officialWorldRecords}
           />
           <Box sx={{ mt: 4 }}>
             <FormControlLabel
--- a/client/src/components/admin/RoundDoubleCheck/RoundDoubleCheck.js
+++ b/client/src/components/admin/RoundDoubleCheck/RoundDoubleCheck.js
@@ -48,6 +48,13 @@ const ROUND_QUERY = gql`
         enteredAt
       }
     }
+    officialWorldRecords {
+      event {
+        id
+      }
+      type
+      attemptResult
+    }
   }
 `;
 
@@ -103,7 +110,7 @@ function RoundDoubleCheck() {
 
   if (loading && !data) return <Loading />;
   if (error) return <Error error={error} />;
-  const { round } = data;
+  const { round, officialWorldRecords } = data;
 
   const results = orderBy(
     round.results,
@@ -164,6 +171,7 @@ function RoundDoubleCheck() {
             cutoff={round.cutoff}
             disabled={enterLoading}
             onSubmit={handleResultAttemptsSubmit}
+            officialWorldRecords={officialWorldRecords}
           />
         </Grid>
         <Grid item md sx={{ textAlign: 'center' }}>
--- a/client/src/lib/attempt-result.js
+++ b/client/src/lib/attempt-result.js
@@ -264,9 +264,8 @@ export function isWorldRecord(
   attemptResult,
   eventId,
   type,
-  officialWorldRecords
+  officialWorldRecords = []
 ) {
-  // if (officialWorldRecords === undefined) return false;
   const wr =
     officialWorldRecords.find(
       (wr) => wr.type === type && wr.event.id === eventId
--- a/client/src/lib/attempt-result.js
+++ b/client/src/lib/attempt-result.js
@@ -257,10 +257,32 @@ export function autocompleteTimeAttemptResult(time) {
 }
 
 /**
+ * Checks whether a given attempt is a world record of the given type.
+ * Returns the corresponding boolean.
+ */
+export function isWorldRecord(
+  attemptResult,
+  eventId,
+  type,
+  officialWorldRecords
+) {
+  // if (officialWorldRecords === undefined) return false;
+  const wr =
+    officialWorldRecords.find(
+      (wr) => wr.type === type && wr.event.id === eventId
+    ) || null;
+  return wr !== null && attemptResult <= wr.attemptResult;
+}
+
+/**
  * Checks the given attempt results for discrepancies and returns
  * a warning message if some are found.
  */
-export function attemptResultsWarning(attemptResults, eventId) {
+export function attemptResultsWarning(
+  attemptResults,
+  eventId,
+  officialWorldRecords
+) {
   const skippedGapIndex =
     trimTrailingSkipped(attemptResults).indexOf(SKIPPED_VALUE);
   if (skippedGapIndex !== -1) {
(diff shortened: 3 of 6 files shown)


client/src/lib/attempt-result.js
```diff
@@ -256,18 +256,71 @@ export function autocompleteTimeAttemptResult(time) {
...
+    eventId,
+    officialWorldRecords
+  );
+  if (newWorldRecordSingle) {
+    return `
+      The result you're trying to submit includes a new world record single
+      (${formatAttemptResult(bestSingle, eventId)}).
+      Please check that the results are accurate.
+    `;
+  }
```
jonatanklosko: We want to wrap this in `if (completeAttempts.length > 0)`.

client/src/lib/attempt-result.js
```diff
@@ -256,18 +256,71 @@ export function autocompleteTimeAttemptResult(time) {
...
 
+/**
+ * Checks if a given set of results is a new world record average and returns a corresponding boolean.
+ */
+export function isWorldRecordAverage(
+  attemptResults,
+  eventId,
+  officialWorldRecords
+) {
+  const newAverage = average(attemptResults, eventId);
```
jonatanklosko: We calculate the average already outside this function, so instead i would do `isWorldRecord(attemptResult, eventId, type, officialWorldRecords)`. Where `attemptResult` is either single or average and it's given by `type`.

client/src/lib/attempt-result.js
```diff
@@ -256,18 +256,71 @@ export function autocompleteTimeAttemptResult(time) {
...
+  const wrAverage = officialWorldRecords?.find(
+    (wr) => wr.type === 'average' && wr.event.id === eventId
+  )?.attemptResult;
+  // if no wrAverage is found for some reason (networking issue, brand new event, etc.),
+  // we get `number < undefined` which always evaluates to `false`
```
jonatanklosko: We should always have the records, they are fetched together with the round, so if there are networking issues then we wouldn't load the page in the first place :)

client/src/components/admin/AdminRound/AdminRoundContent.js
```diff
@@ -135,6 +135,7 @@ function AdminRoundContent({ round, competitionId }) {
             focusOnResultChange={true}
             disabled={loading}
             onSubmit={handleResultAttemptsSubmit}
+            officialWorldRecords={officialWorldRecords}
           />
```
jonatanklosko, line 139: We also use the form in `RoundDoubleCheck`, so please update it accordingly :)

client/src/lib/attempt-result.js
nsilvestri: I moved that line to a higher scope so now it includes the `333mbf` branch. Given my understanding of how the code works, I don't think this should affect behavior.

client/src/lib/attempt-result.js
nsilvestri: With that in mind, I removed the comment altogether

client/src/components/admin/AdminRound/AdminRoundContent.js
nsilvestri: Oops! I checked if `AdminRoundContent` had been used anywhere else but not the form. Fixed!

client/src/lib/attempt-result.js
nsilvestri: I think I understand what you are saying. Please check the latest commit and see if that was the implementation and interface you were thinking of.

client/src/components/admin/RoundDoubleCheck/RoundDoubleCheck.js
```diff
@@ -48,6 +48,13 @@ const ROUND_QUERY = gql`
         enteredAt
       }
     }
+    officialWorldRecords {
+      event {
+        id
+      }
+      type
+      attemptResult
+    }
```
nsilvestri, line 57: Just copied over this part of the query from `AdminRound.js`. Not sure if I should box this up into a shared fragment. Is that what a fragment is? I don't know GQL terminology :P

client/src/lib/attempt-result.js
```diff
@@ -256,36 +256,82 @@ export function autocompleteTimeAttemptResult(time) {
...
+export function isWorldRecord(
+  attemptResult,
+  eventId,
+  type,
+  officialWorldRecords
+) {
+  const wr = officialWorldRecords?.find(
+    (wr) => wr.type === type && wr.event.id === eventId
+  )?.attemptResult;
+  return attemptResult <= wr;
```
jonatanklosko, line 272:
```suggestion
  const wr = officialWorldRecords.find(
    (wr) => wr.type === type && wr.event.id === eventId
  ) || null;
  return wr !== null && attemptResult <= wr.attemptResult;
```

client/src/lib/tests/attempt-result.test.js
```diff
@@ -295,6 +296,72 @@ describe('meetsCutoff', () => {
...
+          id: '333',
+        },
+        attemptResult: 200,
+      },
+    ];
+    expect(isWorldRecord(attemptResults, '333', 'single', worldRecords)).toBe(
+      false
+    );
+  });
+});
```
jonatanklosko, line 373: Let's add one more tests when there is no world record entry of the given type (MLBD average), in which case it should return `false`.

client/src/lib/attempt-result.js
```suggestion
  const wr = officialWorldRecords.find(
    (wr) => wr.type === type && wr.event.id === eventId
  ) || null;
  return wr !== null && attemptResult <= wr.attemptResult;
```

client/src/components/admin/RoundDoubleCheck/RoundDoubleCheck.js
jonatanklosko: In this case I think it's fine without the fragment, it's brief and we use it only in two places.

jonatanklosko:
Awesome, two small comments and we can ship it :)

client/src/lib/attempt-result.js
nsilvestri: Suggested code was complaining when `officialWorldRecords === undefined` so I added a check in `isWorldRecord` that returns false when it is not defined.

nsilvestri: Existing tests caused `officialWorldRecords` to sometimes be undefined, so I added an empty array in those tests where the WR status of an attempt was not relevant to the test.

client/src/lib/tests/attempt-result.test.js
nsilvestri: Done!

client/src/lib/attempt-result.js
```diff
@@ -256,36 +256,84 @@ export function autocompleteTimeAttemptResult(time) {
...
+ * Checks whether a given attempt is a world record of the given type.
+ * Returns the corresponding boolean.
+ */
+export function isWorldRecord(
+  attemptResult,
+  eventId,
+  type,
+  officialWorldRecords
+) {
+  // if (officialWorldRecords === undefined) return false;
```
```suggestion
  officialWorldRecords = []
) {
```

client/src/lib/attempt-result.js
```diff
@@ -256,36 +256,83 @@ export function autocompleteTimeAttemptResult(time) {
...
+
 /**
  * Checks the given attempt results for discrepancies and returns
  * a warning message if some are found.
  */
-export function attemptResultsWarning(attemptResults, eventId) {
+export function attemptResultsWarning(
+  attemptResults,
+  eventId,
+  officialWorldRecords
```
```suggestion
  officialWorldRecords = []
```

jonatanklosko (approved):
Perfect, thank you!

PR #168, opened 2023-08-10
Replace button prop with ListItemButton

maxidragon:
Button prop is deprecated: https://mui.com/material-ui/api/list-item/#ListItem-prop-button


jonatanklosko (approved):
Thanks :)

PR #169, opened 2023-08-10
Display number of participated competitors

### LINKED ISSUE #59: Number of participated competitors
We need some record with number of presented competitors (who have at least one event). It can be useful for media announcements and WCA Dues counting).
Maybe here https://live.worldcubeassociation.org/admin/competitions/<COMPETITION_ID>/competitors

maxidragon:
Closes #59

### DIFF
--- a/client/src/components/admin/AdminCompetitors/AdminCompetitors.jsx
+++ b/client/src/components/admin/AdminCompetitors/AdminCompetitors.jsx
@@ -1,6 +1,6 @@
 import { gql, useQuery } from "@apollo/client";
 import { useParams } from "react-router-dom";
-import { Typography } from "@mui/material";
+import { Grid, Typography } from "@mui/material";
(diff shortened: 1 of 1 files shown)


client/src/components/admin/AdminCompetitors/AdminCompetitors.jsx
```diff
@@ -50,11 +50,22 @@ function AdminCompetitors() {
...
+    )
+  ).length;
 
   return (
     <>
       <Typography variant="h5" gutterBottom>
-        Competitors
+        Competitors ({participatedCompetitors} /{" "}
+        {competition.competitors.length})
       </Typography>
```
jonatanklosko: This maybe be a bit misleading, if the number is here I would expect it to be the same as the number of competitors in the table.

Instead, I suggest we put this information below the table. I'd show something like "Total: {...}" and "Competed: {...}" :)

jonatanklosko:
Thanks!

PR #179, opened 2023-08-17
Added WPA/BPA display for final rounds

### LINKED ISSUE #174: Add live statistics (only for finals rounds?)
Would be great if WCA live provided statistics that make it easier to track a round during live scoretaking. 
* Solve 1: Rank by single
* Solve 2: Rank by mean of 2 solves
* Solve 3 - 5: Display WPA/BPA, provide sorting options to user to rank by either WPA or BPA?

maxidragon:
Closes #174 

Now final round looks like this:

image

Projector view:

image

I'm not sure how sorting should be resolved, I think it's not bad like this, but I'm open to suggestions on how it could be done better.


jonatanklosko:
Hey, thanks for the PR! A few high level notes:

1. I think green/red calls too much attention. Instead, let's use regular color but with opacity (such that it's clearly visible, but feels like a secondary detail).
2. For rounds with mo3 we can show the current mean after 2 solves. Particularly useful for FMC.
3. I don't think we need the toggle for now. With the opacity it should be subtle enough. If it turns out confusing to people, we can have a settings icon with more control.
4. It would be nice to show a tooltip on each of these extra numbers with brief explanation of what the number is.

maxidragon:
I made changes ;) 

Now it looks like this:
image
image
image
image
image

dunkOnIT:
Thanks, this looks awesome! One small nit - it looks like WPA / BPA are off-centre in your second-last screenshot - could we investigate the reason for this?

maxidragon:
@dunkOnIT fixed ;)

image

client/src/components/RoundResults/RoundResultsTable.jsx
```diff
@@ -130,7 +137,59 @@ const RoundResultsTable = memo(
...
+                                  >
+                                    {calculateBpa(
+                                      result.attempts.map(
+                                        (attempt) => attempt.result
+                                      ),
+                                      eventId
+                                    )}
+                                  </Typography>
+                                </Tooltip>
+                                {" / "}
```
jonatanklosko: This should get opacity too :)

client/src/components/RoundResults/RoundResultsTable.jsx
```diff
@@ -130,7 +137,59 @@ const RoundResultsTable = memo(
                     }}
                   >
                     <RecordTagBadge litePr recordTag={result[recordTagField]}>
-                      {formatAttemptResult(result[field], eventId)}
+                      {result.average === 0 && field === "average"
+                        ? roundFormat.id === "a"
+                          ? result.attempts.length > 3 && (
+                              <>
```
jonatanklosko: Let's move this to a separate component, this way we should be able to make the conditional branches easier to follow. I think we can actually have a component that works both here and in the results projector, it's fine to define the component in this file and export.

client/src/components/RoundResults/RoundResultsTable.jsx
```diff
@@ -130,7 +137,59 @@ const RoundResultsTable = memo(
...
+                                <Tooltip title="Best possible average">
+                                  <Typography
+                                    variant="body2"
+                                    component="span"
+                                    sx={{ opacity: 0.5 }}
+                                  >
```
jonatanklosko: We don't need typography here, the relevant styles come from the table cell. The main difference between this one and projector is the font-size, which we wouldn't have to pass otherwise :) For opacity you can do `<Box component="span" sx={{ opacity: 0.5 }}>` (a regular `span` with `style` works fine too).

client/src/lib/attempt-result.js
```diff
@@ -197,6 +197,54 @@ export function formatAttemptResult(attemptResult, eventId) {
   return centisecondsToClockFormat(attemptResult);
 }
 
+/**
+ * Calcualte BPA (Best Possible Average) for the given attempts.
+ * Returns a human-friendly string.
```
jonatanklosko: For consistency with other calculation functions, let's return the numeric value and format the value in the component. This way we decouple the actual logic from presentation :)

client/src/lib/attempt-result.js
```diff
@@ -197,6 +197,54 @@ export function formatAttemptResult(attemptResult, eventId) {
...
+ * @example
+ * calculateBpa([6111, 6000, 5999, 6000], '333'); // => '1:00.00'
+ * calculateBpa([6111, -1, -1, 6000], '333'); // => 'DNF'
+ * calculateBpa([6111, -1, 6000, 5999], '333'); // => '1:00.36'
+ */
+export function calculateBpa(times, eventId) {
```
jonatanklosko: Nitpick: it's generally a good idea to avoid acronyms unless they are very well establish. So I'd do this: `bestPossibleAverage`, `worstPossibleAverage`, `incompleteMean`.

Also note that the average needs to be rounded accordingly, and for FMC we want to multiply the attempts by 100 to represent 2 decimal places in the integer. See the implementation of `average`, we can closely match the rules from there.

jonatanklosko (commented):
This looks awesome! A couple comments with respect to code :)

client/src/components/RoundResults/RoundResultsTable.jsx
```diff
@@ -130,7 +131,12 @@ const RoundResultsTable = memo(
                     }}
                   >
                     <RecordTagBadge litePr recordTag={result[recordTagField]}>
-                      {formatAttemptResult(result[field], eventId)}
+                      <Result
```
jonatanklosko: `Result` -> `ResultStat` to be more specific

client/src/lib/attempt-result.js
```diff
@@ -197,6 +197,56 @@ export function formatAttemptResult(attemptResult, eventId) {
...
+/**
+ * Calculate BPA (Best Possible Average) for the given attempts.
+ * @example
+ * bestPossibleAverage([3642, 3102, 3001, 2992]); // => 3032
+ * bestPossibleAverage([6111, -1, -1, 6000]); // => -1
+ * bestPossibleAverage([4822, 4523, 4233, -1]; // => 4526
+ */
+export function bestPossibleAverage(times) {
+  let bpa;
+  console.log(times);
```
```suggestion
```

client/src/components/Result/Result.jsx
```diff
@@ -0,0 +1,58 @@
+import { Box, Tooltip } from "@mui/material";
+import {
+  bestPossibleAverage,
+  worstPossibleAverage,
+  formatAttemptResult,
+  incompleteMean,
+} from "../../lib/attempt-result";
+
+const Result = ({ result, field, eventId, roundFormat }) => {
```
jonatanklosko:
```suggestion
function Result({ result, field, eventId, roundFormat }) {
```

client/src/components/Result/Result.jsx
```diff
@@ -0,0 +1,58 @@
...
+  return (
+    <>
+      {result.average === 0 && field === "average"
+        ? roundFormat.id === "a"
+          ? result.attempts.length > 3 && (
```
jonatanklosko: Now that it's a separate component, we can use regular `if`/`else` to improve readability, something along these lines:

```jsx
if (result.average === 0 && field === "average") {
  if (roundFormat.id === "a") {
    return ...
  }

  ...
}

return ...
```

jonatanklosko (approved):
Fantastic, thanks :)

PR #195, opened 2023-10-01
Add API for entering individual attempts

jonatanklosko:
Introduces personal competition-specific scoretaking tokens and an API endpoint for entering an individual attempt result. For more details see [the wiki page](https://github.com/thewca/wca-live/wiki/Entering-attempts-with-external-devices).

### DIFF
--- a/lib/wca_live/accounts/scoretaking_token.ex
+++ b/lib/wca_live/accounts/scoretaking_token.ex
@@ -16,7 +16,9 @@ defmodule WcaLive.Accounts.ScoretakingToken do
   @validity_in_days 7
 
   schema "scoretaking_tokens" do
-    field :token, :string
+    # We only store the token hash, as we would do with password
+    field :token, :string, virtual: true
+    field :token_hash, :binary
 
     belongs_to :user, Accounts.User
     belongs_to :competition, Competitions.Competition
@@ -25,12 +27,27 @@ defmodule WcaLive.Accounts.ScoretakingToken do
   end
 
   @doc """
+  Computes hash for the given token.
+  """
+  @spec token_hash(String.t()) :: binary()
+  def token_hash(token) do
+    :crypto.hash(:sha256, token)
+  end
+
+  @doc """
   Generates a new competition-scoped token for the given user.
   """
-  @spec build_scoretaking_token(%Accounts.User{}, %Competitions.Competition{}) :: %__MODULE__{}
+  @spec build_scoretaking_token(%Accounts.User{}, %Competitions.Competition{}) ::
+          {String.t(), %__MODULE__{}}
   def build_scoretaking_token(user, competition) do
     token = :crypto.strong_rand_bytes(@rand_size) |> Base.url_encode64(padding: false)
-    %ScoretakingToken{token: token, user_id: user.id, competition_id: competition.id}
+
+    %ScoretakingToken{
+      token: token,
+      token_hash: token_hash(token),
+      user_id: user.id,
+      competition_id: competition.id
+    }
   end
 
   @doc """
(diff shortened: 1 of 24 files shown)


lib/wca_live/accounts/scoretaking_token.ex
```diff
@@ -0,0 +1,70 @@
...
+      select: {user, competition}
+    )
+  end
+
+  @doc """
+  Returns query to lookup token struct for the given token value.
+  """
+  @spec token_query(String.t()) :: Ecto.Query.t()
+  def token_query(token) do
+    from(ScoretakingToken, where: [token: ^token])
```
jfly: If I'm reading this correctly, it looks like we're storing the API tokens in plaintext. To me, this feels like storing passwords in plaintext (someone with a dump of your database would be able to turn around and make API calls to the live site). Just going off of my personal experience, websites seems to be of 2 minds about this:


- I'm guessing that GitHub hashes their personal access tokens, because they emphasize that you only get to see it once.
image
- Datadog definitely stores their API keys in plaintext, you can get back to them anytime:
image
- I wasn't able to find a great resource about by Googling (searching for "storing api tokens" turns up a lot of results about people who are writing *clients* for various apis). I did find [this opinionated SO answer](https://security.stackexchange.com/questions/180345/do-i-need-to-hash-or-encrypt-api-keys-before-storing-them-in-a-database#:~:text=Yes%2C%20you%20should%20absolutely%20hash,note%20that's%20hashed%20%2D%20not%20encrypted.) that says you should "absolutely" hash API keys, but it doesn't get into the nuance of the different user experience it requires.


For the record, I'm not suggesting a redesign. I'm just curious how much thought you put into this? If you did, it would be nice to add a comment or update your commit message with thoughts on why you opted to do things this way.

jfly (commented):
(I only skimmed this, I don't feel comfortable giving a shipit)

This is super cool! I'm really excited to see the applications people build one top of this. (I quit my job back in 2015 to work on automated scoretaking, and found that we had to implement [a whole results platform first](https://github.com/jfly/ccm), which of course never actually got finished for a whole host of reasons.)

(totally not a blocker, but) Have you thought at all about access tokens for an organization/party, that is, something not tied to a specific human being?

lib/wca_live/accounts/scoretaking_token.ex
jonatanklosko: Hey, this is a very good call! The difference from passwords is that tokens usually have expiration of some kind (in our case they are competition-specific, so the surface area is pretty small), and also removing all tokens is more of an option. The main usability tradeoff is whether we want to always make the token accessible to the user, and also hashing the token for lookup on every request (but that's not really a big price). All that to say, there is really no reason not to hash, so I will change it :)

jfly: > The difference from passwords is that tokens usually have expiration of some kind (in our case they are competition-specific, so the surface area is pretty small)

Ah, I didn't realized these were scoped to a competition. I thought these were long-lived secrets that I might bake into my automated scoretaking tool.

> whether we want to always make the token accessible to the user

I agree, this feels like the biggest consideration to me.

> All that to say, there is really no reason not to hash, so I will change it :)

Cool! Sorry to create more work for you.

jonatanklosko: Haha, glad to change it now rather than once it's deployed :D

jonatanklosko:
> (totally not a blocker, but) Have you thought at all about access tokens for an organization/party, that is, something not tied to a specific human being?

I forgot to reply. Currently we track which user entered each result (or more precisely made the last edit), so a personal token is easiest to maintain this behaviour. With external devices this information isn't necessarily as meaningful anymore, but if this turns out to be the direction, we will likely need more changes to better accommodate this use case. I went with the simplest viable solution, that allows people to test "scorecardless" projects without hacking around some bot-based data entry (first such competition is expected next month) :)

PR #209, opened 2024-02-10
Competition records list

### LINKED ISSUE #58: Competition records list
_From Diana Murnik who make posts about Russian NRs in our media_

It will be convenient if we can see all the records of selected competition. Maybe some separate tab?

maxidragon:
Closes #58 

image


test/wca_live/scoretaking_test.exs
```diff
@@ -801,6 +801,21 @@ defmodule WcaLive.ScoretakingTest do
...
+    competition_event = insert(:competition_event, competition: competition)
+    another_competition_event = insert(:competition_event, competition: another_competition)
+    round = insert(:round, competition_event: competition_event)
+    another_round = insert(:round, competition_event: another_competition_event)
+    record = insert(:result, round: round, best: 300, average: 550, single_record_tag: "WR")
+    insert(:result, round: another_round, best: 300, average: 550, average_record_tag: "WR")
+
+    competition_records = Scoretaking.list_competition_records(competition)
+    assert 1 = length(competition_records)
+    assert record.id == Enum.at(competition_records, 0).result.id
```
jonatanklosko: We can use the existing helper to make it a bit more declarative, like this:

```suggestion
    result =
      insert_result_x_days_ago(
        5,
        single_record_tag: "WR",
        average_record_tag: nil,
        best: 300,
        average: 550
      )

    _other_competition_result =
      insert_result_x_days_ago(
        5,
        single_record_tag: nil,
        average_record_tag: "WR",
        best: 300,
        average: 550
      )
    
    competition = result.round.competition_event.competition

    assert [single_wr] = Scoretaking.list_competition_records(competition)

    assert result.id == single_wr.result.id
    assert :single = single_wr.type
    assert "WR" = single_wr.tag
    assert 300 = single_wr.attempt_result
```

client/src/components/CompetitionNavigation/CompetitionNavigation.jsx
```diff
@@ -59,6 +60,7 @@ function CompetitionNavigation() {
         <Route path="rounds/:roundId/*" element={<Round />} />
         <Route path="competitors" element={<Competitors />} />
         <Route path="competitors/:competitorId" element={<Competitor />} />
+        <Route path="records" element={<CompetitionRecords />} />
```
jonatanklosko: I think we should put it directly on the competition page under schedule, but display the section only if there is any record. Otherwise people may click on the icon just to find out there are no records, and also having it on the main page makes the record stand out, which makes sense.

jonatanklosko:
Thanks for the PR, let's move the section and I think it's good :)

maxidragon:
Now it looks like this:

image

client/src/components/CompetitionHome/CompetitionHome.jsx
```diff
@@ -18,13 +18,44 @@ import { wcaUrl } from "../../lib/urls";
...
 
 const COMPETITION_QUERY = gql`
   query Competition($id: ID!) {
     competition(id: $id) {
       id
       wcaId
       name
+      competitionRecords {
+        id
+        tag
```
jonatanklosko: Let's create `components/RecordList/fragments.jsx` with `RECORD_LIST_RECORD_FRAGMENT`, so that we can share it in both places.

Existing example:

https://github.com/thewca/wca-live/blob/f11c7634e9778f8bd4098908a138f90800541669/client/src/components/admin/AdminRound/fragments.jsx#L5-L6

jonatanklosko (approved):

jonatanklosko:
Deployed, see [example](https://live.worldcubeassociation.org/competitions/4382) :)

PR #215, opened 2024-03-27
Added checks for regulation 9m

### LINKED ISSUE #186: Check that there are enough competitors to start the next round before opening a round
There was recently a case where WCA Live allowed a 4th round to be opened despite the first round only having 90 people. This led to some results needing to be removed after the competition as it was not realized at the competition that this 4th round was not legal ([9m1](https://www.worldcubeassociation.org/regulations/#9m1)). A check for this would ensure that this situation would not happen again.

mckeenicholas:
Closes #186

I've only implemented the check on the server side as of now. I can add a client side check as well if necessary.


OldManLink:
There are still grammar mistakes, BTW: "rounds with less than N competitors..." should be "with fewer than N competitors..."

Also, just a suggestion, why not use the actual rule text which is subtly different, even if semantically the same as yours:

```
9m1) Rounds with 99 or fewer competitors must have at most two subsequent rounds.
9m2) Rounds with 15 or fewer competitors must have at most one subsequent round.
9m3) Rounds with 7 or fewer competitors must not have subsequent rounds.
```

mckeenicholas:
I've updated the text to use the regulation text. The original code had a check for having at least 8 competitors in the previous round which was formatted in the same way I wrote them, which was my reasoning for keeping it in the same format. (I also find "fewer than 100" more understandable than "99 or fewer", but its probably better to use the actual regulation text).

lib/wca_live/scoretaking.ex
```diff
@@ -224,41 +224,61 @@ defmodule WcaLive.Scoretaking do
...
 
+  # Check that enough people have competed in the previous round in order to open the next round
+  # See https://www.worldcubeassociation.org/regulations/#9m
+  defp validate_previous_rounds(round) do
+    min = [8, 16, 100]
+    round_results =
+      Enum.to_list(1..round.number - 1 // 1)
+      |> Enum.map(
+        fn round_offset ->
+          Round |> Round.where_sibling(round, -round_offset) |> Repo.one() |> Repo.preload(:results)
```
jonatanklosko, line 260: Instead of querying each round separately, let's add `list_previous_rounds`. It should return rounds ordered by round number, and here we can reverse for the check.

lib/wca_live/scoretaking.ex
```diff
@@ -224,41 +224,61 @@ defmodule WcaLive.Scoretaking do
...
+        end
+      )
+      |> Enum.map(fn round -> round.results |> Enum.reject(&Result.empty?/1) |> length() end)
+
+    Enum.find_index(Enum.zip(round_results, min), fn {x, y} -> x < y end)
+    |> case do
+      0 -> "Rounds with 7 or fewer competitors must not have subsequent rounds"
+      1 -> "Rounds with 15 or fewer competitors must have at most one subsequent round"
+      2 -> "Rounds with 99 or fewer competitors must have at most two subsequent rounds"
+      _ -> :nil
```
jonatanklosko, line 270: The only other value we expected is nil (not found), so let's be explicit:

```suggestion
      nil -> nil
```

(also, `nil` is an atom, but you don't need `:` because it's special, same with `true` and `false` :))

lib/wca_live/scoretaking.ex
```diff
@@ -224,41 +224,61 @@ defmodule WcaLive.Scoretaking do
...
+          Round |> Round.where_sibling(round, -round_offset) |> Repo.one() |> Repo.preload(:results)
+        end
+      )
+      |> Enum.map(fn round -> round.results |> Enum.reject(&Result.empty?/1) |> length() end)
+
+    Enum.find_index(Enum.zip(round_results, min), fn {x, y} -> x < y end)
+    |> case do
+      0 -> "Rounds with 7 or fewer competitors must not have subsequent rounds"
+      1 -> "Rounds with 15 or fewer competitors must have at most one subsequent round"
+      2 -> "Rounds with 99 or fewer competitors must have at most two subsequent rounds"
```
jonatanklosko, line 269: Usually we keep `{:error, _}` messages lowercased. We transform to uppercase in other places as needed.

```suggestion
      0 -> "rounds with 7 or fewer competitors must not have subsequent rounds"
      1 -> "rounds with 15 or fewer competitors must have at most one subsequent round"
      2 -> "rounds with 99 or fewer competitors must have at most two subsequent rounds"
```

lib/wca_live/scoretaking.ex
```diff
@@ -224,41 +224,61 @@ defmodule WcaLive.Scoretaking do
...
     else
-      Multi.new()
-      |> Multi.run(:finish_previous, fn _repo, _changes ->
-        if round.number > 1 do
-          previous = get_previous_round(round) |> Repo.preload(:results)
-          finish_round(previous)
-        else
-          {:ok, nil}
+      message = validate_previous_rounds(round)
+      if message do
```
jonatanklosko, line 228: To avoid nesting you can do this:

```elixir
cond do
  Round.open?(round) ->
    {:error, ...}

  message = validate_previous_rounds(round) ->
    {:error, error}

  true ->
    Multi.new()
    |> ...
end
```

jonatanklosko:
@mckeenicholas thanks, a couple small comments. Also, please add one more test similar to this one:

https://github.com/thewca/wca-live/blob/66b1ab6e0d78617ce957d94725a4994aaa0efa71/test/wca_live/scoretaking_test.exs#L524-L534

PR #216, opened 2024-03-28
Make numpad '+' and '-' mimic up and down arrows

### LINKED ISSUE #99: Feature Request: Allow numpad '+' and '-' to be used as down and up arrows and add a "delete mode"
As the '+' and '-' keys on the numpad are unused, it would be convenient to allow those keys to be used as up and down arrow keys while entering results. Another suggestion would tobe allow '.' on the numpad to be used as backspace as it is not required while entering results, however, that may be confusing to those who are unaware.

My second suggestion is to add a "delete mode" in order to easily remove no-shows. While it is enabled, you can simply type in a competitor id and press enter to delete them from the round. This is much faster than the current system of searching the list of competitors in the round, and allows scoretakers to quickly go through empty leftover scorecards of no-shows.

mckeenicholas:
Closes #99


jonatanklosko (approved):
@mckeenicholas looks good, thanks!

PR #225, opened 2024-08-02
Introduce distinct markers for results with uncertain advancment

### LINKED ISSUE #128: Make it clearer when advancing competitors have been finalized
It's not obvious to a new competitor that the advancing competitors are not finalized until all results for the round are in. A few ideas:
* Display a disclaimer saying that advancements are not final until all results are in.
* Display the percentage of results of the round have been entered.
* Make the advancing ranks yellow instead of green until all results for the round have been entered.

### LINKED ISSUE #222: Distinction between "will advance" and "might advance"
It can be a bit confusing when competitors' positions are marked green, prior to the whole round being completed. This makes some new competitors believe they have advanced, just because they competed in group 1.

Suggestion:
* if a competitor did advance (even if everyone left to enter either dropped out, for the 75% rule, or beat them, they would still advance), then use the current dark green color
* if a competitor might advance (if everyone left is slower than them, then they would advance), use a lighter green

jonatanklosko:
Closes #222. Also closes #128.

An ongoing round looks like this:

image

Ranks in the darker green "clinched advancement", while the ones in lighter green qualify as is, but may not advance depending on the remaining results.

I reverted #111 (ignoring empty results in percentage calculation), since those "uncertain" results will now show in the light color.

This change introduces an additional flag to results specifically for the display purpose, so it doesn't affect any of the existing logic.

### DIFF
--- a/lib/wca_live/scoretaking/advancing.ex
+++ b/lib/wca_live/scoretaking/advancing.ex
@@ -22,10 +22,15 @@ defmodule WcaLive.Scoretaking.Advancing do
   @spec compute_advancing(%Round{}) :: Ecto.Changeset.t(%Round{})
   def compute_advancing(round) do
     round = round |> Repo.preload(:results)
-    advancing = advancing_results(round)
+    {advancing, clinched_advancing} = advancing_results(round)
 
     Enum.map(round.results, fn result ->
-      Changeset.change(result, advancing: result in advancing)
+      advancing = result in advancing
+
+      Changeset.change(result,
+        advancing: advancing,
+        advancing_questionable: advancing and result not in clinched_advancing
+      )
     end)
     |> Round.put_results_in_round(round)
   end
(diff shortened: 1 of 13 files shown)


lib/wca_live/scoretaking/advancing.ex
```diff
@@ -210,6 +212,32 @@ defmodule WcaLive.Scoretaking.Advancing do
     end)
   end
 
+  defp clinched_qualifying_results(round) do
+    # Assume best possible attempts for empty results and see who of
+    # the currently entered results would still qualify.
```
jonatanklosko, line 217: FTR we don't need to worry about results dropping out, because getting a better result than everybody is a "stronger assumption". Intuitively, when someone gets a better result than everybody they make 1 person not advance, but when they drop out they effectively make 0.75 person not advance.

PR #236, opened 2024-10-03
Add confirmation dialog if there is a DNS followed by a valid attempt

maxidragon:


client/src/lib/attempt-result.js
```diff
@@ -483,3 +489,15 @@ export function meetsCutoff(attemptResults, cutoff) {
...
+/**
+ * Checks if the given attempt results contain a DNS followed by a valid result.
+ */
+export function checkForDnsFollowedByValidResult(attemptResults) {
+  const dnsIndex = attemptResults.findIndex((attempt) => attempt === DNS_VALUE);
+  if (dnsIndex === -1) return false;
+  const validResultIndex = attemptResults.findIndex(
+    (attempt, index) => (attempt > 0 || attempt === -1) && index > dnsIndex
+  );
+  return validResultIndex !== -1;
```
```suggestion
  return attemptResults.some(
    (attempt, index) => index > dnsIndex && (attempt !== SKIPPED_VALUE && attempt !== DNS_VALUE)
  );
```

Both checks work, but I think this is a bit more clear about the intent.

client/src/lib/attempt-result.js
```diff
@@ -483,3 +489,15 @@ export function meetsCutoff(attemptResults, cutoff) {
     .slice(0, numberOfAttempts)
     .some((attempt) => attempt > 0 && attempt < attemptResult);
 }
+
+/**
+ * Checks if the given attempt results contain a DNS followed by a valid result.
+ */
+export function checkForDnsFollowedByValidResult(attemptResults) {
```
jonatanklosko: Currently this is just an implementation detail of the warning function, so I would keep it private. If the idea is to export it for testing, usually it's better to test the caller (here, the warning function).

```suggestion
function checkForDnsFollowedByValidResult(attemptResults) {
```

client/src/lib/tests/attempt-result.test.js
```diff
@@ -538,6 +548,23 @@ describe("applyTimeLimit", () => {
...
+    const attemptResults = [1000, DNF_VALUE, 2500, DNF_VALUE, 2000];
+    expect(checkForDnsFollowedByValidResult(attemptResults)).toEqual(false);
+  });
+
+  it("returns false if there is no valid result after DNS", () => {
+    const attemptResults = [1000, DNF_VALUE, DNF_VALUE, DNS_VALUE, DNS_VALUE];
+    expect(checkForDnsFollowedByValidResult(attemptResults)).toEqual(false);
+  });
+});
+
```
jonatanklosko: The warning test is enough, it already covers the happy paths too.

```suggestion
```

jonatanklosko (approved):
Good call :)

PR #237, opened 2024-10-07
Show the number of entered results in the events overview

### LINKED ISSUE #228: Overview for scoretaking progress
In admin view being able to get an overview over how many results have been entered in each round could be helpful. 

Perhaps something like this:
image

This would help getting an overview of what is done and what is missing. 

Alternatively some sort of indicator on whether a round is fully entered or not.

maxidragon:
Closes #228 

Not sure how to style this - any ideas? It looks like this:

image

### DIFF
--- a/client/src/components/admin/AdminCompetitionEvents/AdminCompetitionEvents.jsx
+++ b/client/src/components/admin/AdminCompetitionEvents/AdminCompetitionEvents.jsx
@@ -21,6 +21,8 @@ const COMPETITION_EVENTS_QUERY = gql`
           name
           open
           finished
+          enteredResults
+          totalResults
         }
       }
     }
--- a/client/src/components/admin/AdminCompetitionEvents/AdminCompetitionEvents.jsx
+++ b/client/src/components/admin/AdminCompetitionEvents/AdminCompetitionEvents.jsx
@@ -21,8 +21,8 @@ const COMPETITION_EVENTS_QUERY = gql`
           name
           open
           finished
-          enteredResults
-          totalResults
+          numEnteredResults
+          numResults
         }
       }
     }
--- a/client/src/components/admin/AdminCompetitionEvents/ClearRoundButton.jsx
+++ b/client/src/components/admin/AdminCompetitionEvents/ClearRoundButton.jsx
@@ -9,6 +9,8 @@ const CLEAR_ROUND_MUTATION = gql`
       round {
         id
         open
+        enteredResults
+        totalResults
       }
     }
   }
--- a/client/src/components/admin/AdminCompetitionEvents/ClearRoundButton.jsx
+++ b/client/src/components/admin/AdminCompetitionEvents/ClearRoundButton.jsx
@@ -9,8 +9,8 @@ const CLEAR_ROUND_MUTATION = gql`
       round {
         id
         open
-        enteredResults
-        totalResults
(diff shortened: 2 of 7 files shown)


jonatanklosko:
I updated the design slightly:

image

We already use the "N of M entered" wording in the round view, so I used the same for consistency. I think it looks fine, but we can always revisit.

client/src/components/admin/AdminCompetitionEvents/AdminCompetitionEvents.jsx
```diff
@@ -21,6 +21,8 @@ const COMPETITION_EVENTS_QUERY = gql`
           name
           open
           finished
+          enteredResults
+          totalResults
```
jonatanklosko: Reading this names without context one may assume the values is a list of results (technically we don't have a nested `{}`, but that's a detail). I would make the naming more explicit, as in `numEnteredResults` and `numResults`.

client/src/components/admin/AdminCompetitionEvents/ClearRoundButton.jsx
```diff
@@ -9,6 +9,8 @@ const CLEAR_ROUND_MUTATION = gql`
       round {
         id
         open
+        enteredResults
+        totalResults
```
jonatanklosko: I added this for both open and clear. The reason for including particular fields in the graphql response is that the local cache and therefore existing queries are immediately updated. Without this, after opening a round we would see `0 of 0 entered` :)

maxidragon:
It looks much better!

lib/wca_live/scoretaking/round.ex
```diff
@@ -134,6 +134,23 @@ defmodule WcaLive.Scoretaking.Round do
...
+
+  def entered_results(round) do
+    format = Format.get_by_id!(round.format_id)
+
+    entered_results =
+      round.results
+      |> Enum.filter(&Result.has_expected_attempts?(&1, format.number_of_attempts, round.cutoff))
+      |> length()
+
+    entered_results
```
jonatanklosko:
```suggestion
    
    Enum.count(round.results, &Result.has_expected_attempts?(&1, format.number_of_attempts, round.cutoff))
```

lib/wca_live/scoretaking/round.ex
```diff
@@ -134,6 +134,23 @@ defmodule WcaLive.Scoretaking.Round do
...
+  end
+
+  def total_results(round) do
+    length(round.results)
+  end
```
jonatanklosko: For this one, I would just do `length(round.results)` directly :)

```suggestion
```

lib/wca_live/scoretaking/round.ex
```diff
@@ -134,6 +134,17 @@ defmodule WcaLive.Scoretaking.Round do
       (unfinished_results / total_results < 0.1 and not active?(round))
   end
 
+  def num_entered_results(%Round{results: []}), do: 0
+
+  def num_entered_results(round) do
```
jonatanklosko:
```suggestion
  @doc """
  Returns the number of results that have all the expected attempts entered.
  """
  @spec num_entered_results(%Round{}) :: non_neg_integer()
  def num_entered_results(round) do
```

jonatanklosko (approved):
Thanks!

PR #240, opened 2024-11-22
new bulk quit experience

ticklemepierce:
Allows bulk quitting to be done via keyboard only (but can still be done with a mouse/trackpad if one wishes)

### DIFF
--- a/client/package-lock.json
+++ b/client/package-lock.json
@@ -14,6 +14,7 @@
         "@cubing/icons": "^1.0.6",
         "@emotion/react": "^11.5.0",
         "@emotion/styled": "^11.3.0",
+        "@mui/base": "^5.0.0-beta.62",
         "@mui/icons-material": "^5.1.0",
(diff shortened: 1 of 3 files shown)


jonatanklosko:
Hey @ticklemepierce, thanks for the PR :) I like the idea. I experimented with multi-select and I like the result, see #242. The user can search by name or id (same as the result form). The select automatically provides more affordances, such as removing selections with backspace, and selecting with up/down arrows + enter.

ticklemepierce:
> Hey @ticklemepierce, thanks for the PR :) I like the idea. I experimented with multi-select and I like the result, see #242. The user can search by name or id (same as the result form). The select automatically provides more affordances, such as removing selections with backspace, and selecting with up/down arrows + enter.

Yeah I just wasn't sure if that would be too much of a UI change which is why I did it how I did :P but I think I initially tried something like this haha

PR #241, opened 2024-11-22
Require the user to type a confirmation keyword when submitting a world record

ticklemepierce:
require data entry to type "world record" for world records since people have clicked through the modal still

### DIFF
--- a/client/src/components/admin/ResultAttemptsForm/ResultAttemptsForm.jsx
+++ b/client/src/components/admin/ResultAttemptsForm/ResultAttemptsForm.jsx
@@ -96,17 +96,16 @@ function ResultAttemptsForm({
   }
 
   function confirmSubmission() {
-    const { description, confirmationKeyword } = attemptResultsWarning(
+    const submissionWarning = attemptResultsWarning(
       attemptResults,
       eventId,
       officialWorldRecords
     );
 
-    if (description) {
+    if (submissionWarning) {
       return confirm({
-        description,
+        ...submissionWarning,
         confirmationText: "Submit",
-        confirmationKeyword,
       });
     } else {
       return Promise.resolve();
--- a/client/src/components/admin/ResultAttemptsForm/ResultAttemptsForm.jsx
+++ b/client/src/components/admin/ResultAttemptsForm/ResultAttemptsForm.jsx
@@ -96,15 +96,15 @@ function ResultAttemptsForm({
   }
 
   function confirmSubmission() {
-    const [submissionWarning, confirmationKeyword] = attemptResultsWarning(
+    const { description, confirmationKeyword } = attemptResultsWarning(
       attemptResults,
       eventId,
       officialWorldRecords
     );
 
-    if (submissionWarning) {
+    if (description) {
       return confirm({
-        description: submissionWarning,
+        description,
         confirmationText: "Submit",
         confirmationKeyword,
       });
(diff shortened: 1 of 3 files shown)


maxidragon:
I'm not sure if we need something like this. I think that just clicking "Confirm" is enough. However, let's wait for Jonatan's opinion

ticklemepierce:
People have just clicked through it still though.  The idea is to take a page from github and AWS's books and require people to be much more explicit and slow them down.

jonatanklosko:
Hey! Given that world records are rare, I think an extra confirmation doesn't hurt :)

client/src/components/admin/ResultAttemptsForm/ResultAttemptsForm.jsx
```diff
@@ -96,7 +96,7 @@ function ResultAttemptsForm({
   }
 
   function confirmSubmission() {
-    const submissionWarning = attemptResultsWarning(
+    const [submissionWarning, confirmationKeyword] = attemptResultsWarning(
```
jonatanklosko: Given that the second element is optional, let's make it return an object, so it's more structured. It would be either `{ text: "...", confirmationKeyword: "..." }` or `{ text: "..." }`.

```suggestion
    const submissionWarning = attemptResultsWarning(
```

client/src/components/admin/ResultAttemptsForm/ResultAttemptsForm.jsx
```diff
@@ -106,6 +106,7 @@ function ResultAttemptsForm({
       return confirm({
         description: submissionWarning,
         confirmationText: "Submit",
+        confirmationKeyword,
```
jonatanklosko:
```suggestion
        description: submissionWarning.text,
        confirmationText: "Submit",
        confirmationKeyword: submissionWarning.confirmationKeyword,
```

client/src/lib/attempt-result.js
```diff
@@ -364,9 +364,9 @@ export function attemptResultsWarning(
   const skippedGapIndex =
     trimTrailingSkipped(attemptResults).indexOf(SKIPPED_VALUE);
   if (skippedGapIndex !== -1) {
-    return `You've omitted attempt ${
+    return [`You've omitted attempt ${
```
jonatanklosko: Please change to an object as suggested above. Note that we also need to update tests in `attempt-result.test.js` :)

ticklemepierce:
Yeah I didn't love doing it as an array so object makes sense - admittedly was pretty tired when I made the PR lol.  Will update in a bit

client/src/components/admin/ResultAttemptsForm/ResultAttemptsForm.jsx
ticklemepierce: just noting that I used `description` instead of `text` so that it can just be
```
{
  description,
  confimationText: "Submit",
  confirmationKeyword,
}
```

by destructuring on line 99.  Then the conditional checking for `submissionWarning` can be changed to `description` since that is analogous to the previous functionality.

client/src/lib/attempt-result.js
```diff
@@ -414,29 +420,29 @@ export function attemptResultsWarning(
...
+            There's a big difference between the best single
+            (${formatAttemptResult(bestSingle, eventId)}) and the worst single
+            (${formatAttemptResult(worstSingle, eventId)}).
+            Please check that the results are accurate.`
+        };
       }
     }
   }
-  return null;
+  return {};
```
jonatanklosko: I think returning an empty object is not meaningful here, so I would prefer to go with the approach I originally suggested :)

jonatanklosko: Destructuring could still be done `const {...} = attemptResultsWarning(...) || {}`, though I wouldn't do that either, since `if (submissionWarning)` is more clear.

jonatanklosko (approved):

PR #243, opened 2024-11-25
Add API endpoint to export competition results

coder13:
It'd be useful to get results periodically as they're being added. This would enable a number of third party apps that make use of live results.

### DIFF
--- a/lib/wca_live/competitions.ex
+++ b/lib/wca_live/competitions.ex
@@ -58,8 +58,13 @@ defmodule WcaLive.Competitions do
   @spec get_competition_by_wca_id!(String.t()) :: %Competition{}
   def get_competition_by_wca_id!(wca_id), do: Repo.get_by!(Competition, wca_id: wca_id)
 
-  @spec find_competition_by_wca_id!(String.t()) :: %Competition{}
-  def find_competition_by_wca_id!(wca_id), do: Repo.get_by(Competition, wca_id: wca_id)
+  @doc """
+  Gets a single competition by human-readable WCA id.
+
+  Does not raise an error if no competition is found.
+  """
+  @spec find_competition_by_wca_id(String.t()) :: %Competition{}
+  def find_competition_by_wca_id(wca_id), do: Repo.get_by(Competition, wca_id: wca_id)
 
   @doc """
   Gets a single competition.
--- a/lib/wca_live/competitions.ex
+++ b/lib/wca_live/competitions.ex
@@ -60,11 +60,9 @@ defmodule WcaLive.Competitions do
 
   @doc """
   Gets a single competition by human-readable WCA id.
-
-  Does not raise an error if no competition is found.
   """
-  @spec find_competition_by_wca_id(String.t()) :: %Competition{}
-  def find_competition_by_wca_id(wca_id), do: Repo.get_by(Competition, wca_id: wca_id)
+  @spec get_competition_by_wca_id(String.t()) :: %Competition{}
+  def get_competition_by_wca_id(wca_id), do: Repo.get_by(Competition, wca_id: wca_id)
 
   @doc """
   Gets a single competition.
--- a/lib/wca_live/competitions.ex
+++ b/lib/wca_live/competitions.ex
@@ -71,14 +71,7 @@ defmodule WcaLive.Competitions do
   @spec fetch_competition_by_wca_id(String.t()) ::
           {:ok, %Competition{}} | {:error, Ecto.Queryable.t()}
   def fetch_competition_by_wca_id(id) do
-    Repo.fetch(where(Competition, wca_id: ^wca_id))
-  end
-
-  defp parse_competition_id_or_wca_id(id) do
-    case Integer.parse(id, 10) do
-      {competition_id, ""} -> {:competition_id, competition_id}
-      _ -> {:wca_id, id}
-    end
+    Repo.fetch(where(Competition, wca_id: ^id))
   end
 
   @doc """
--- a/lib/wca_live/competitions.ex
+++ b/lib/wca_live/competitions.ex
@@ -58,17 +58,19 @@ defmodule WcaLive.Competitions do
   @spec get_competition_by_wca_id!(String.t()) :: %Competition{}
   def get_competition_by_wca_id!(wca_id), do: Repo.get_by!(Competition, wca_id: wca_id)
 
+  @spec fetch_competition(any()) :: {:error, any()} | {:ok, %{optional(atom()) => any()}}
   @doc """
-  Gets a single competition by human-readable WCA id.
+  Gets a single competition
   """
-  @spec get_competition_by_wca_id(String.t()) :: %Competition{}
-  def get_competition_by_wca_id(wca_id), do: Repo.get_by(Competition, wca_id: wca_id)
+  @spec fetch_competition(term()) :: {:ok, %Competition{}} | {:error, Ecto.Queryable.t()}
+  def fetch_competition(id), do: Repo.fetch(Competition, id)
 
   @doc """
   Gets a single competition by numeric id or wca id
   """
-  @spec fetch_competition(term()) :: {:ok, %Competition{}} | {:error, Ecto.Queryable.t()}
-  def fetch_competition(id) do
+  @spec fetch_competition_by_id_or_wca_id(term()) ::
+          {:ok, %Competition{}} | {:error, Ecto.Queryable.t()}
+  def fetch_competition_by_id_or_wca_id(id) do
     case parse_competition_id_or_wca_id(id) do
       {:competition_id, competition_id} -> Repo.fetch(Competition, competition_id)
       {:wca_id, wca_id} -> Repo.fetch(Ecto.Query.where(Competition, wca_id: ^wca_id))
@@ -76,7 +78,7 @@ defmodule WcaLive.Competitions do
   end
 
   defp parse_competition_id_or_wca_id(id) do
-    case Integer.parse(id) do
+    case Integer.parse(id, 10) do
       {competition_id, ""} -> {:competition_id, competition_id}
       _ -> {:wca_id, id}
     end
--- a/lib/wca_live/competitions.ex
+++ b/lib/wca_live/competitions.ex
@@ -66,15 +66,12 @@ defmodule WcaLive.Competitions do
   def fetch_competition(id), do: Repo.fetch(Competition, id)
 
   @doc """
-  Gets a single competition by numeric id or wca id
+  Gets a single competition by WCA id.
   """
-  @spec fetch_competition_by_id_or_wca_id(term()) ::
+  @spec fetch_competition_by_wca_id(String.t()) ::
           {:ok, %Competition{}} | {:error, Ecto.Queryable.t()}
-  def fetch_competition_by_id_or_wca_id(id) do
-    case parse_competition_id_or_wca_id(id) do
-      {:competition_id, competition_id} -> Repo.fetch(Competition, competition_id)
-      {:wca_id, wca_id} -> Repo.fetch(Ecto.Query.where(Competition, wca_id: ^wca_id))
-    end
+  def fetch_competition_by_wca_id(id) do
+    Repo.fetch(where(Competition, wca_id: ^wca_id))
   end
 
   defp parse_competition_id_or_wca_id(id) do
--- a/lib/wca_live/competitions.ex
+++ b/lib/wca_live/competitions.ex
@@ -45,7 +45,7 @@ defmodule WcaLive.Competitions do
   end
 
   @doc """
-  Gets a single competition.
+  Gets a single competition by numeric id.
   """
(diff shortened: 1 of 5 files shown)


lib/wca_live_web/router.ex
```diff
@@ -48,6 +48,8 @@ defmodule WcaLiveWeb.Router do
     pipe_through :api
 
     get "/competitions/:id/wcif", CompetitionController, :show_wcif
+    get "/competitions/:id/results", CompetitionController, :show_results
+    get "/competitions/:wca_id/id", CompetitionController, :show_competition_id
```
jonatanklosko:
```suggestion

    # Public results endpoint
    get "/competitions/:id_or_wca_id/results", CompetitionController, :show_results
```

If the idea is to request the id just to make the results request, I would rather make it `/competitions/:id_or_wca_id/results`. We can do a similar logic as here:

https://github.com/thewca/wca-live/blob/7c01c5a94b04ddad49758c5b441d2be1c683ec69/lib/wca_live_web/controllers/link_controller.ex#L27-L44

We can determine which id we are dealing with, because one is integer, and the other is not.

lib/wca_live/competitions.ex
```diff
@@ -58,6 +58,14 @@ defmodule WcaLive.Competitions do
   @spec get_competition_by_wca_id!(String.t()) :: %Competition{}
   def get_competition_by_wca_id!(wca_id), do: Repo.get_by!(Competition, wca_id: wca_id)
 
+  @doc """
+  Gets a single competition by human-readable WCA id.
+
+  Does not raise an error if no competition is found.
+  """
+  @spec find_competition_by_wca_id(String.t()) :: %Competition{}
+  def find_competition_by_wca_id(wca_id), do: Repo.get_by(Competition, wca_id: wca_id)
```
jonatanklosko:
```suggestion
  @doc """
  Gets a single competition by human-readable WCA id.
  """
  @spec get_competition_by_wca_id(String.t()) :: %Competition{}
  def get_competition_by_wca_id(wca_id), do: Repo.get_by(Competition, wca_id: wca_id)
```

lib/wca_live/synchronization/export.ex
```diff
@@ -13,6 +13,12 @@ defmodule WcaLive.Synchronization.Export do
     |> competition_to_wcif()
   end
 
+  def export_results(competition) do
```
jonatanklosko: The `Synchronization.{Import,Export}` modules are specifically about WCIF and I wouldn't mix the responsibilities.

I think it's actually fine to build the response in the controller (normally we would have a separate "view" modules for JSON responses, but for a single endpoint either way is fine).

coder13: I can refactor for this.

lib/wca_live/synchronization/export.ex
```diff
@@ -208,4 +214,37 @@ defmodule WcaLive.Synchronization.Export do
...
+      "rounds" => competition_event.rounds |> Enum.map(&round_for_results/1)
+    }
+  end
+
+  defp round_for_results(round) do
+    %{
+      "id" => round_wcif_id(round),
+      "roundNumber" => round.number,
+      "cutoff" => round.cutoff && round.cutoff |> cutoff_to_wcif(),
+      "timeLimit" => round.time_limit && round.time_limit |> time_limit_to_wcif(),
```
jonatanklosko: I wouldn't include cutoff/timelimit. I'm thinking the response could be this:

```json
{
  "rounds": [
    {
      "eventId": "333",
      "number": 1,
      "results": [
        {
          "personId": 1,
          "best": 100,
          "average": 100,
          "attemptResults": [100, 100, 100, 100, 100]
        }
      ]
    }
  ],
  "persons": [
    {
      "id": 1,
      "wcaId": "2000FOOO01",
      "name": "Foo Bar"
    }
  ]
}
```

It's worth extracting the person data, so there is less duplication.

jonatanklosko:
Note that we also need tests in `competition_controller_test.exs`.

coder13:
I've added a test

Here is the current output: https://gist.github.com/coder13/826c4fdf0dab0d589d05e93caae3bd19

I think it makes sense to include ranking.

I wasn't sure if you meant to flatten the array of rounds but in the spirit of reducing duplication, I group rounds by event.

I return all people, I don't know the simplest solution for returning only people with results but that would also make sense as an optimization.

I played around with filtering out people without results / empty rounds / events with no rounds with results

lib/wca_live_web/controllers/competition_controller.ex
```diff
@@ -26,6 +27,25 @@ defmodule WcaLiveWeb.CompetitionController do
     end
   end
 
+  @spec show_results(Plug.Conn.t(), nil | maybe_improper_list() | map()) :: Plug.Conn.t()
+  def show_results(conn, params) do
```
jonatanklosko:
```suggestion
  def show_results(conn, params) do
```

lib/wca_live_web/controllers/competition_controller.ex
```diff
@@ -26,6 +27,25 @@ defmodule WcaLiveWeb.CompetitionController do
...
+    IO.inspect(params)
+
+    case Competitions.fetch_competition_by_id_or_wca_id(params["id_or_wca_id"]) do
```
jonatanklosko:
```suggestion
    case Competitions.fetch_competition_by_id_or_wca_id(params["id_or_wca_id"]) do
```

lib/wca_live_web/controllers/competition_controller.ex
```diff
@@ -174,4 +194,64 @@ defmodule WcaLiveWeb.CompetitionController do
         {:error, "the provided token is not valid"}
     end
   end
+
+  @spec format_results(%Competition{}) :: map()
+  def format_results(competition) do
```
jonatanklosko:
```suggestion
  defp format_results(competition) do
```

lib/wca_live_web/controllers/competition_controller.ex
```diff
@@ -174,4 +194,64 @@ defmodule WcaLiveWeb.CompetitionController do
...
+    competition
+    |> Repo.preload(
+      competition_events: [rounds: [:competition_event, results: [:person]]],
+      people: []
+    )
+    |> competition_to_results()
+  end
+
+  defp competition_to_results(competition) do
```
jonatanklosko, line 210: To only include the people with results, do this:

```elixir
result_registrant_ids =
  for competition_event <- competition.competition_events,
      round <- competition_event.rounds,
      result <- round.results,
      do: result.registrant_id,
      into: MapSet.new()
```

And then:

```elixir
competition.people
|> Enum.filter(& &1.registrant_id in result_registrant_ids)
|> ...
```

lib/wca_live_web/controllers/competition_controller.ex
```diff
@@ -174,4 +194,64 @@ defmodule WcaLiveWeb.CompetitionController do
...
+    %{
+      events:
+        competition.competition_events
+        |> Enum.map(&format_competition_event/1)
+        |> Enum.filter(&(&1["rounds"] != [])),
```
jonatanklosko: I think it's fine to include all events and also rounds.

test/wca_live_web/controllers/competition_controller_test.exs
```diff
@@ -17,6 +17,48 @@ defmodule WcaLiveWeb.CompetitionControllerTest do
       assert %{"formatVersion" => "1.0", "id" => "WC2019"} = body
     end
 
+    test "returns results for a competition", %{conn: conn} do
+      competition = insert(:competition, wca_id: "WC2019")
+      competition_event = insert(:competition_event, competition: competition)
+      round = insert(:round, competition_event: competition_event)
+      person = insert(:person, competition: competition)
+      result = insert(:result, round: round, person: person)
```
jonatanklosko, line 46: Let's add person without results to make sure they are not returned :)

```suggestion
      result = insert(:result, round: round, person: person)
      _person_without_result = insert(:person, competition: competition)
```

lib/wca_live/competitions.ex
```diff
@@ -58,12 +58,32 @@ defmodule WcaLive.Competitions do
...
+  Gets a single competition by numeric id or wca id
+  """
+  @spec fetch_competition_by_id_or_wca_id(term()) ::
+          {:ok, %Competition{}} | {:error, Ecto.Queryable.t()}
+  def fetch_competition_by_id_or_wca_id(id) do
+    case parse_competition_id_or_wca_id(id) do
+      {:competition_id, competition_id} -> Repo.fetch(Competition, competition_id)
+      {:wca_id, wca_id} -> Repo.fetch(Ecto.Query.where(Competition, wca_id: ^wca_id))
+    end
+  end
```
jonatanklosko: I would do the parsing and conditional call in controller. We already have `fetch_competition` above, so we just need this:

```suggestion
  @doc """
  Gets a single competition by WCA id.
  """
  @spec fetch_competition_by_wca_id(String.t()) ::
          {:ok, %Competition{}} | {:error, Ecto.Queryable.t()}
  def fetch_competition_by_wca_id(id) do
    Repo.fetch(where(Competition, wca_id: ^wca_id))
  end
```

The main reason is that I don't like passing an "unstructured" id string and parsing it here.

lib/wca_live_web/controllers/competition_controller.ex
```diff
@@ -174,4 +190,76 @@ defmodule WcaLiveWeb.CompetitionController do
...
+      for competition_event <- competition.competition_events,
+          round <- competition_event.rounds,
+          result <- round.results,
+          do: result.person.registrant_id,
+          into: MapSet.new()
+
+    %{
+      events:
+        competition.competition_events
+        |> Enum.map(&format_competition_event/1),
```
jonatanklosko:
```suggestion
      events: Enum.map(competition.competition_events, &format_competition_event/1),
```

lib/wca_live_web/controllers/competition_controller.ex
```diff
@@ -174,4 +190,76 @@ defmodule WcaLiveWeb.CompetitionController do
...
+      persons:
+        competition.people
+        |> Enum.filter(&(&1.registrant_id in result_registrant_ids))
+        |> Enum.map(&format_person/1)
+        |> Enum.filter(&(&1["id"] != nil))
```
jonatanklosko: Just a detail, but each filter/map will do a separate pass on the list, we can make it into a single pass with the comprehension syntax:

```suggestion
      persons:
        for person <- competition.people, person.registrant_id in result_registrant_ids do
          format_person(person)
        end
```

Also, I don't think we need to filter id != nil. I don't remember where the registrant id would be nil, but people that have results definitely do have the id.

lib/wca_live_web/controllers/competition_controller.ex
```diff
@@ -174,4 +190,76 @@ defmodule WcaLiveWeb.CompetitionController do
...
+    }
+  end
+
+  defp format_competition_event(competition_event) do
+    %{
+      "eventId" => competition_event.event_id,
+      "rounds" =>
+        competition_event.rounds
+        |> Enum.map(&format_round/1)
```
jonatanklosko:
```suggestion
      "rounds" => Enum.map(competition_event.rounds, &format_round/1)
```

test/wca_live_web/controllers/competition_controller_test.exs
```diff
@@ -17,6 +17,49 @@ defmodule WcaLiveWeb.CompetitionControllerTest do
       assert %{"formatVersion" => "1.0", "id" => "WC2019"} = body
     end
 
+    test "returns results for a competition", %{conn: conn} do
```
jonatanklosko, line 41: Btw. currently this is in the `"show_wcif"` describe block. Let's move it to its own describe block below :)

lib/wca_live_web/controllers/competition_controller.ex
```diff
@@ -174,4 +190,76 @@ defmodule WcaLiveWeb.CompetitionController do
...
+    }
+  end
+
+  defp format_round(round) do
+    %{
+      "number" => round.number,
+      "results" =>
+        round.results |> Enum.map(&format_result/1) |> Enum.filter(&(&1["attempts"] != []))
```
jonatanklosko: Again, just an optimisation to do a single pass (and filtering upfront is nicer :D).

```suggestion
        for result <- round.results, result.attempts != [] do
          format_result(result)
        end
```

jonatanklosko (commented):
Just nitpicks and it should be good to go :)

lib/wca_live_web/controllers/competition_controller.ex
coder13: I did this because I'd get non-competing users like "PNW Cubing", the users we use to direct invoices to the org. For some reason they'd have nil IDs.

image

But with your suggestion, they already get filtered out with `|> Enum.filter(&(&1.registrant_id in result_registrant_ids))`

jonatanklosko (approved):

PR #245, opened 2024-12-04
Add page for anonymizing personal data

jonatanklosko:
Streamlines anonymization to not require synchronizing individual competitions.


jonatanklosko:
After opening the PR, I realised that at the person's first competition wouldn't get anonymized, because WCA Live wouldn't have a WCA ID for the person at that point, in which case synchronization is still the way to go.

Instead I reduced the timespan that old competitions are kept to 90 days (#246).

PR #248, opened 2024-12-26
Handle leading/trailing spaces in search input

maxidragon:
It's very annoying when you paste a competition name that accidentally has a space before the actual name, causing the search to fail. The same issue occurs if there are, for example, two spaces between words. 

After these changes, it's no longer possible to start a search with a space or type multiple consecutive spaces. I couldn't find a better solution to this, if anyone has suggestions feel free to comment on this PR.

### DIFF
--- a/test/wca_live/competitions_test.exs
+++ b/test/wca_live/competitions_test.exs
@@ -28,6 +28,7 @@ defmodule WcaLive.CompetitionsTest do
     assert worlds.id in ids(list)
     assert open.id not in ids(list)
     assert euro.id in ids(list)
+
     list = Competitions.list_competitions(%{filter: " championship  2022  "})
 
     assert worlds.id in ids(list)
--- a/test/wca_live/competitions_test.exs
+++ b/test/wca_live/competitions_test.exs
@@ -28,26 +28,7 @@ defmodule WcaLive.CompetitionsTest do
     assert worlds.id in ids(list)
     assert open.id not in ids(list)
     assert euro.id in ids(list)
-  end
-
-  test "list_competitions/1 given :filter with leading space matches competitions by name" do
-    worlds = insert(:competition, name: "World Championship 2020")
-    open = insert(:competition, name: "WHAT Open 2020")
-    euro = insert(:competition, name: "Euro Championship 2020")
-
-    list = Competitions.list_competitions(%{filter: " World"})
-
-    assert worlds.id in ids(list)
-    assert open.id not in ids(list)
-    assert euro.id not in ids(list)
-  end
-
-  test "list_competitions/1 given :filter with trailing space matches competitions by name" do
-    worlds = insert(:competition, name: "World Championship 2020")
-    open = insert(:competition, name: "WHAT Open 2020")
-    euro = insert(:competition, name: "Euro Championship 2020")
-
-    list = Competitions.list_competitions(%{filter: "Championship  "})
+    list = Competitions.list_competitions(%{filter: " championship  2022  "})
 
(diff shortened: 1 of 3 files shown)


jonatanklosko:
We could only change the value send in the request, but I would actually move this all the way up and make the server decide. It would be here: https://github.com/thewca/wca-live/blob/5f546f5137105b86bf0652282c6bc0467193ba3a/lib/wca_live/competitions.ex#L39-L41

You can add

```ex
filter = filter |> String.trim() |> String.replace(~r/\s+/, " ")
```

and then we can expand this test:

https://github.com/thewca/wca-live/blob/5f546f5137105b86bf0652282c6bc0467193ba3a/test/wca_live/competitions_test.exs#L21-L31

maxidragon:
That would be nice, but I've tried trimming the name on the backend (and changing the value only for the request), and it won't actually work because the `Autocomplete` component does its own filtering, regardless of what is in the array.

jonatanklosko:
We can skip the default `Autocomplete` filtering, by adding this:

```js
filterOptions={(options, state) => options}
```

maxidragon:
> We can skip the default `Autocomplete` filtering, by adding this:
> 
> ```js
> filterOptions={(options, state) => options}
> ```

This is exactly what I was looking for but couldn’t find, thanks!

test/wca_live/competitions_test.exs
```diff
@@ -30,6 +30,30 @@ defmodule WcaLive.CompetitionsTest do
...
+    assert open.id not in ids(list)
+    assert euro.id not in ids(list)
+  end
+
+  test "list_competitions/1 given :filter with trailing space matches competitions by name" do
+    worlds = insert(:competition, name: "World Championship 2020")
+    open = insert(:competition, name: "WHAT Open 2020")
+    euro = insert(:competition, name: "Euro Championship 2020")
+
+    list = Competitions.list_competitions(%{filter: "Championship  "})
```
jonatanklosko: I would just add this to test all whitespace at once :D

```suggestion
    list = Competitions.list_competitions(%{filter: " championship  2022  "})
```

test/wca_live/competitions_test.exs
```diff
@@ -25,6 +25,11 @@ defmodule WcaLive.CompetitionsTest do
 
     list = Competitions.list_competitions(%{filter: "champion"})
 
+    assert worlds.id in ids(list)
+    assert open.id not in ids(list)
+    assert euro.id in ids(list)
+    list = Competitions.list_competitions(%{filter: " championship  2022  "})
```
jonatanklosko:
```suggestion
    assert euro.id in ids(list)

    list = Competitions.list_competitions(%{filter: " championship  2022  "})
```

jonatanklosko (approved):

test/wca_live/competitions_test.exs
```diff
@@ -28,6 +28,12 @@ defmodule WcaLive.CompetitionsTest do
     assert worlds.id in ids(list)
     assert open.id not in ids(list)
     assert euro.id in ids(list)
+
+    list = Competitions.list_competitions(%{filter: " championship  2022  "})
```
jonatanklosko:
```suggestion
    list = Competitions.list_competitions(%{filter: " championship  2020  "})
```

jonatanklosko:
Thanks!

test/wca_live/competitions_test.exs
maxidragon: I was about to commit the same fix, but github returned an error that you had already made the change 😅

PR #252, opened 2025-01-17
Forecast view

HaysCubing:
"Forecast View" calculates projected averages for all competitors and then ranks the results based on projected average.

Projected average is shown for incomplete results.


client/src/components/Round/Round.jsx
```diff
@@ -72,6 +76,12 @@ const ROUND_UPDATED_SUBSCRIPTION = gql`
   ${ROUND_RESULT_FRAGMENT}
 `;
 
+// Events sorted by best don't need forecast view.
+// Fewest moves is currently unsupported
+export function forecastViewDisabled(format, eventId) {
```
HaysCubing: Is there a cleaner way to set up "forecastViewDisabled"? I want it to be disabled (icon and functionality) for "Best of" events and FMC. It looks like I need to pass the const "forecastView" since that's what's initialized with useState() and causes the page to reload. However calling forecastViewDisabled() within every child element generator seems inefficient.

client/src/components/Round/RoundToolbar.jsx
```diff
@@ -23,6 +25,16 @@ function RoundToolbar({ round, competitionId }) {
       <Grid item style={{ flexGrow: 1 }} />
       {mdScreen && (
         <Grid item>
+          {!forecastViewDisabled(round.format, round.competitionEvent.event.id) &&
```
HaysCubing: It's an icon for right now. Should it be a checkbox?

client/vite.config.js
```diff
@@ -15,5 +15,7 @@ export default defineConfig({
     proxy: {
       "/auth": "http://127.0.0.1:4000",
     },
+    watch: {usePolling: true},
```
HaysCubing: Remove this before submitting

client/src/components/Round/Round.jsx
HaysCubing: Cleaned this up a bit. No longer passing to child functions. If you navigate from a page with it enabled (ex. 7x7) to a page disabled (BLD), and forecast view is enabled, then line 125 will disable it and do a page reload. So loads the page twice essentially - not sure if this is a big deal or not

client/src/components/Round/RoundToolbar.jsx
HaysCubing: Made it a checkbox. More clear this way. Right now I have it so the entire thing doesn't appear at all if it's not supported. I could also just have it disabled for non supported events. Is that better?

client/src/components/Round/Round.jsx
```diff
@@ -110,12 +121,16 @@ function Round() {
   if (error) return <Error error={error} />;
   const { round } = data;
 
+  if (forecastView && forecastViewDisabled(round)) {
+    setForecastView(false);
+  }
```
jonatanklosko: I believe `set*` functions should not be called in the function body directly, but rather in response to some events. Here it may actually work because with the `if` makes sure it doesn't get called over and over, but I would do this instead:

```suggestion
  useEffect(() => {
    // Reset to default on round change
    setForecastView(false);
  }, [roundId]);
```

client/src/components/Round/Round.jsx
```diff
@@ -72,6 +76,12 @@ const ROUND_UPDATED_SUBSCRIPTION = gql`
   ${ROUND_RESULT_FRAGMENT}
 `;
 
+// Events sorted by best don't need forecast view.
+// Fewest moves is currently unsupported
+export function forecastViewDisabled(round) {
+  return round.format.sortBy === "best" || round.competitionEvent.event.id === "333fm";
+}
```
jonatanklosko: Is there a reason to exclude FMC? We only need to `* 100` the values when computing average, as in the existing `average` function.

client/src/components/Round/RoundToolbar.jsx
```diff
@@ -23,6 +27,17 @@ function RoundToolbar({ round, competitionId }) {
...
         <Grid item>
+          {!forecastViewDisabled(round) &&
+            <FormControlLabel
+              control={
+                <Checkbox
+                  checked={forecastView}
+                  onChange={(event) => setForecastView(event.target.checked)}
+                />
+              }
+              label="Forecast View"
```
jonatanklosko: I think this takes too much space, but don't worry about it. I will experiment with a toggle icon before shipping.

client/src/lib/attempt-result.js
```diff
@@ -124,6 +139,107 @@ function mean(values) {
...
+    }
+    if (currentResult.projectedAverage === prevResult.projectedAverage &&
+      currentResult.best === prevResult.best) {
+      // Rankings tie
+      currentResult.ranking = prevResult.ranking;
+    } else {
+      currentResult.ranking = i + 1;
+    }
+    // Using simple advancing logic.
+    // TODO: Implement questionable/clinched logic
```
jonatanklosko: We can actually leave `advancingQuestionable` untouched and it should be accurate. If `advancingQuestionable` is true, it means the result is advancing even if all remaining results are best possible, so it is still valid in the forecast view. If `advancingQuestionable` is false, then it should not become true based on projected average.

It may be worth leaving a comment that this is the case actually.

client/src/lib/attempt-result.js
```diff
@@ -124,6 +139,107 @@ function mean(values) {
   return Math.round(sum / values.length);
 }
 
+/**
+ * return the number of competitors that should be marked as advancing.
+ * When no competitors advance, return 3 for the podium placements.
+ * Might not be accurate for percentage based advancement but that is acceptable
+ */
+function getAdvancingCount(results, advancementCondition) {
```
jonatanklosko: Correct me if I'm wrong, but the point of forecast view is for final rounds, because it's usually those that have results entered partially. So what I would do is enable forecast view only for final rounds, and then for advancing we just mark top 3 ranking.

If a use case appears in the future to support other rounds, I think we should support all advancement conditions correctly.

client/vite.config.js
jonatanklosko:
```suggestion
    // Uncomment when running using Docker on Windows.
    // See https://github.com/vitejs/vite/issues/1153#issuecomment-785467271
    // watch: { usePolling: true },
```

client/src/components/Round/Round.jsx
```diff
@@ -72,6 +76,12 @@ const ROUND_UPDATED_SUBSCRIPTION = gql`
   ${ROUND_RESULT_FRAGMENT}
 `;
 
+// Events sorted by best don't need forecast view.
+// Fewest moves is currently unsupported
+export function forecastViewDisabled(round) {
```
jonatanklosko: Nit: let's move this to `result.js` also to keep the logic together, and let's revert as `forecastViewEnabled` (so that we don't need double negation as in `!forecastViewDisabled(round)`).

client/src/lib/attempt-result.js
```diff
@@ -124,6 +139,107 @@ function mean(values) {
...
+ * forThird: time needed to overtake third place
+ */
+export function getExpandedResults(results, format, forecastView, advancementCondition) {
+  if (results.length == 0 || !forecastView) return results;
+  var expandedResults = results.map((result) => {
+    return {
+      ...result,
+      projectedAverage: computeProjectedAverage(result, format),
+      forFirst: SKIPPED_VALUE,
+      forThird: SKIPPED_VALUE,
```
jonatanklosko: Let's remove for now, unless you are working on adding as part of the PR :)

client/src/lib/attempt-result.js
```diff
@@ -124,6 +139,107 @@ function mean(values) {
...
+  return SKIPPED_VALUE;
+}
+
+/**
+ * return the results object with additional properties:
+ * projectedAverage: projected final average based on current results
+ * forFirst: time needed to overtake first place
+ * forThird: time needed to overtake third place
+ */
+export function getExpandedResults(results, format, forecastView, advancementCondition) {
```
jonatanklosko: Let's move this to `result.js`.

Basically: `attempt-result.js` for various calculations on attempt result (numeric values), and `result.js` for dealing with result objects.

I would also rename this to `resultsForView`.

client/src/lib/attempt-result.js
```diff
@@ -124,6 +139,107 @@ function mean(values) {
...
+  for (let result of expandedResults) {
+    result.advancing = false;
+    result.advancingQuestionable = false;
+  }
+
+  const advancingCount = getAdvancingCount(results, advancementCondition);
+  const roundIncomplete = results.some((result) => isSkipped(result.average));
+
+  // Sort based on projection with tiebreakers on single
+  expandedResults = expandedResults.sort((a, b) => compareProjectedResults(a, b));
```
jonatanklosko: We can simplify a bit, first define this for attempt result:

```js
function toMonotonic(attemptResult) {
  if (isComplete(attemptResult)) {
    return attemptResult;
  } else {
    return Infinity;
  }
}
```

and then here we can do:

```js
expandedResults =
  orderBy(expandedResults, [
    (result) => toMonotonic(result.projectedAverage),
    (result) => toMonotonic(result.best),
  ]);
```

client/src/components/Round/Round.jsx
HaysCubing: Ah yeah the `if` was what made sure it didn't infinitely reload the page. I was looking for something proper like this so thanks for the suggestion. I moved it up next to the other `useEffect` call

client/src/components/Round/Round.jsx
HaysCubing: No reason besides I hadn't gotten around to implementing it yet. Was going to add FMC in a future change but if it's important to support now I can look at that as well

client/src/components/Round/RoundToolbar.jsx
HaysCubing: Ok sounds good. Yeah I wasn't sure about the proper style for the toolbar so if you wanted to handle the UI elements that'd probably be better.

client/src/lib/attempt-result.js
HaysCubing: Wait is that the case for `advancingQuestionable`? What you said I would expect to be true for `advancing`. Isn't `advancingQuestionable` true for all of the competitors currently projected to advance?

This might be obsolete though I change this PR to be just finals

client/src/lib/attempt-result.js
HaysCubing: The extra case here is 3x3 semifinals at worlds. We will be doing live data entry and want to see projections for top 16. I think for now though what you're saying makes more sense. Just support finals and do top 3 highlighting and then I can add semifinals later.

client/src/lib/attempt-result.js
HaysCubing: Oh yeah this is leftover from a more extensive change which has "time needed for X" categories. This is supposed to be in a later PR

client/src/lib/attempt-result.js
HaysCubing: Yeah I just deleted this for now

HaysCubing: Yeah I just deleted this for now

client/src/lib/attempt-result.js
HaysCubing: I made this just finals. I didn't alter `advancing` logic as well.

client/vite.config.js
```diff
@@ -15,5 +15,9 @@ export default defineConfig({
     proxy: {
       "/auth": "http://127.0.0.1:4000",
     },
+    // Uncomment when running using Docker on Windows.
+    // See https://github.com/vitejs/vite/issues/1153#issuecomment-785467271
+    watch: { usePolling: true },
```
HaysCubing, line 20: Comment this out before submitting

client/src/lib/result.js
```diff
@@ -40,3 +47,72 @@ export function paddedAttemptResults(result, numberOfAttempts) {
...
   return padSkipped(attemptResults, numberOfAttempts);
 }
+
+// Events sorted by best don't need forecast view.
+// Fewest moves is currently unsupported
+// Only final rounds are supported
+export function forecastViewEnabled(round) {
+  return round.format.sortBy != "best" &&
+    round.competitionEvent.event.id != "333fm" &&
+    round.name === "Final";
```
HaysCubing: Is this the proper way to check if a round if final? I don't think this will work for "combined final" rounds

client/src/components/RoundResults/RoundResultsTable.jsx
```diff
@@ -47,12 +47,13 @@ const styles = {
...
 
 const RoundResultsTable = memo(
-  ({ results, format, eventId, competitionId, onResultClick }) => {
+  ({ results, format, eventId, competitionId, onResultClick, forecastView }) => {
     const smScreen = useMediaQuery((theme) => theme.breakpoints.up("sm"));
     const mdScreen = useMediaQuery((theme) => theme.breakpoints.up("md"));
 
     const stats = orderedResultStats(eventId, format);
 
+    const expandedResults = resultsForView(results, format, forecastView);
```
jonatanklosko:
```suggestion
    const viewResults = resultsForView(results, format, forecastView);

```

client/src/components/RoundResults/RoundResultsTable.jsx
```diff
@@ -80,7 +81,7 @@ const RoundResultsTable = memo(
             </TableRow>
           </TableHead>
           <TableBody>
-            {results.map((result) => (
+            {expandedResults.map((result) => (
```
jonatanklosko:
```suggestion
            {viewResults.map((result) => (
```

client/src/lib/attempt-result.js
```diff
@@ -124,6 +128,33 @@ function mean(values) {
...
+    }
+    if (attemptResults.length === 3) {
+      const [, x,] = attemptResults.slice().sort(compareAttemptResults);
+      return x;
+    }
+    const [, x, y,] = attemptResults.slice().sort(compareAttemptResults);
+    return meanOfX([x, y]);
+  }
+  return SKIPPED_VALUE;
+}
```
jonatanklosko: Ideally we want to work with plain values in this module, so let's change the function to accept `attemptResults` list as argument. Let's call it `projectedAverage`, since I think `compute` applies all of these functions and is redundant. I would also make the conditional branches more explicit, wdyt?

```suggestion
/**
 * Returns projected average.
 *
 * Note that contrarily to other functions in this module, this
 * function expects a non-padded list of attempt results (without
 * trailing skipped values).
 *
 * Projections are defined as follows:
 *
 *   - mo3 events: mean of current solves
 *   - ao5 events:
 *     - 1-2 solves: mean of current solves
 *     - 3-4 solves: median of current solves
 *
 * When all result attempts are present, the return value is the same
 * as the usual average.
 */
export function projectedAverage(attemptResults, format) {
  if (attemptResults.length === 0) return SKIPPED_VALUE;
  
  if (format.numberOfAttempts === 3) {
    return meanOfX(attemptResults);
  }
  
  if (format.numberOfAttempts === 5) {
    if (attemptResults.length < 3) {
      return meanOfX(attemptResults);
    }
    if (attemptResults.length === 3) {
      const [, x,] = attemptResults.slice().sort(compareAttemptResults);
      return x;
    }
    if (attemptResults.length === 4) {
      const [, x, y,] = attemptResults.slice().sort(compareAttemptResults);
      return meanOfX([x, y]);
    }
    return averageOf5(attemptResults);
  }
  
  throw new Error("Unexpected format");
}
```

client/src/lib/result.js
```diff
@@ -40,3 +47,68 @@ export function paddedAttemptResults(result, numberOfAttempts) {
...
```
jonatanklosko: The easiest way to distinguish is to check there is no advancement condition :D

```suggestion
    round.advancementCondition === null;
```

client/src/lib/result.js
```diff
@@ -40,3 +47,68 @@ export function paddedAttemptResults(result, numberOfAttempts) {
...
+    return {
+      ...result,
+      projectedAverage: computeProjectedAverage(result, format),
+    };
+  });
+
+  if (resultsForView[0].projectedAverage == SKIPPED_VALUE) {
+    // First place has no results. Do nothing.
+    return resultsForView;
+  }
```
jonatanklosko: We need to do the sorting first, otherwise the results with projected average may not be at the top.

client/src/lib/result.js
```diff
@@ -40,3 +47,68 @@ export function paddedAttemptResults(result, numberOfAttempts) {
...
+  ]);
+  resultsForView[0].ranking = 1;
+  var prevResult = resultsForView[0];
+  for (let i = 0; i < resultsForView.length; i++) {
+    let currentResult = resultsForView[i];
+    if (isSkipped(currentResult.projectedAverage)) {
+      break;
+    }
+    if (currentResult.projectedAverage === prevResult.projectedAverage &&
+      currentResult.best === prevResult.best) {
```
jonatanklosko: For comparison we should use `toMonotonic`, so that DNF DNS is treated the same:

```suggestion
    if (toMonotonic(currentResult.projectedAverage) === toMonotonic(prevResult.projectedAverage) &&
      toMonotonic(currentResult.best) === toMonotonic(prevResult.best)) {
```

client/src/lib/result.js
```diff
@@ -40,3 +47,68 @@ export function paddedAttemptResults(result, numberOfAttempts) {
...
+export function resultsForView(results, format, forecastView) {
+  if (results.length == 0 || !forecastView) return results;
+  var resultsForView = results.map((result) => {
+    return {
+      ...result,
+      projectedAverage: computeProjectedAverage(result, format),
+    };
+  });
+
+  if (resultsForView[0].projectedAverage == SKIPPED_VALUE) {
```
jonatanklosko:
```suggestion
  if (isSkipped(resultsForView[0].projectedAverage)) {
```

client/src/lib/result.js
```diff
@@ -40,3 +47,68 @@ export function paddedAttemptResults(result, numberOfAttempts) {
...
+  // Sort based on projection with tiebreakers on single
+  resultsForView = orderBy(resultsForView, [
+    (result) => toMonotonic(result.projectedAverage),
+    (result) => toMonotonic(result.best),
+  ]);
+  resultsForView[0].ranking = 1;
+  var prevResult = resultsForView[0];
+  for (let i = 0; i < resultsForView.length; i++) {
+    let currentResult = resultsForView[i];
+    if (isSkipped(currentResult.projectedAverage)) {
```
jonatanklosko: It's equivalent, but I would check for empty result directly:

```suggestion
    if (currentResult.attempts.length === 0) {
```

client/src/lib/tests/attempt-result.test.js
```diff
@@ -619,3 +620,34 @@ describe("incompleteMean", () => {
...
+    expect(computeProjectedAverage({average: 100}, null)).toEqual(100);
+  });
+
+  var result;
+  it("Returns mean when format is mean of 3", () => {
+    const format = { numberOfAttempts: 3};
+    result = {attempts: [{result: 100}]};
+    expect(computeProjectedAverage(result, format)).toEqual(100);
+    result = {attempts: [{result: 100}, {result: 102}]};
+    expect(computeProjectedAverage(result, format)).toEqual(101);
```
jonatanklosko: Let's test that 0 and 3 attempts also behave as expected (basically we want to test every possible length).

client/src/lib/result.js
```diff
@@ -40,3 +47,68 @@ export function paddedAttemptResults(result, numberOfAttempts) {
...
+      currentResult.best === prevResult.best) {
+      // Rankings tie
+      currentResult.ranking = prevResult.ranking;
+    } else {
+      currentResult.ranking = i + 1;
+    }
+    if (currentResult.ranking <= advancingCount) {
+      // "advancing" field is handled correctly by default results query
+      currentResult.advancingQuestionable = true;
+    }
```
jonatanklosko: I believe the logic we want is this:

```suggestion
    const isClinched = currentResult.advancing && currentResult.advancingQuestionable;
    
    // A clinched result must still be clinched in the projected ranking,
    // so we keep advancing state as is
    if (!isClinched) {
      if (currentResult.ranking <= advancingCount) {
        currentResult.advancing = true;
        currentResult.advancingQuestionable = true;
      } else {
        currentResult.advancing = false;
        currentResult.advancingQuestionable = false;
      }
    }
```

Let's make sure there is a test for the clinched scenario.

client/src/lib/result.js
```diff
@@ -40,3 +47,68 @@ export function paddedAttemptResults(result, numberOfAttempts) {
...
+  });
+
+  if (resultsForView[0].projectedAverage == SKIPPED_VALUE) {
+    // First place has no results. Do nothing.
+    return resultsForView;
+  }
+
+  for (let result of resultsForView) {
+    result.advancingQuestionable = false;
+  }
```
jonatanklosko:
```suggestion
```

(see the clinched suggestion)

client/src/lib/attempt-result.js
jonatanklosko: Gah, I what I said is true for `clinched`, which is kinda reversed. If the result is `clinched` it is still `clinched` in the projected ranking.

`advancingQuestionable` is slightly indirect. I will drop a separate comment on how the conditional would look.

client/src/components/Round/Round.jsx
jonatanklosko: Can totally be a separate PR :)

client/src/lib/result.js
HaysCubing: Technically I think this does work because of how the default ordering is setup and because projectedAverage is always defined if there is at least one attemptResult.

However it relies on ordering from the server and it's cleaner to sort first on the client.

client/src/lib/result.js
HaysCubing: I think I'm confused on how advancing and advancingQuestionable work. It looks like advancingQuestionable doesn't do anything if advancing is set to true?

For this view, I was thinking that we would have the advancing competitors be defined by the query results from the server. Then the new rankings based on projectedAverage would change the advancingQuestionable field since these are based on projected times and so we don't want to indicate they are clinched (advancing = true).

client/src/lib/result.js
HaysCubing: The break statement in this function means that some advancingQuestionable results would not have their value updated based on the new rankings if I make this change.

client/src/lib/attempt-result.js
HaysCubing: Agree on the name change and the conditional branch formatting.

The issue I see here is that for complete results, we are duplicating calculations already done on the server. While they will be the same, I think we should avoid this since it opens up the ability for results to get mismatched.

Explicitly returning result.average when it is complete ensures accuracy for completed results. Since the "projectedAverage" value is something that only exists on the client it makes sense for it to be calculated there.

HaysCubing: I also agree though that it makes more sense for all functions in this module to accept `attemptResults` and not a full `Result` object. Should I create another helper function in result.js that checks explicitly for result.Average, and then calls projectedAverage if it's not already set?

Can also update projectedAverage to indicate that it expects an incomplete data set.

client/src/lib/result.js
HaysCubing: Actually it looks like for clinched results `advancing=true` but `advancingQuestionable=false`? It looks like the formatting for the questionable property also override so that if both are true then the result has the light highlighting.

Wouldn't clinched then be `currentResult.advancing && !currentResult.advancingQuestionable`

client/src/lib/attempt-result.js
HaysCubing: Made the suggested changes and then added a function resultProjectedAverage in result.js that will directly return the average if complete

client/src/lib/result.js
HaysCubing: I updated it so that `advancingQuestionable` isn't altered if `advancing` is set (since questionable overrides default). Otherwise I think it should work as is?

client/src/lib/result.js
HaysCubing: Ok I see now. `clinched = currentResult.advancing && !currentResult.advancingQuestionable`. Therefore I need access to the original advancingQuestionable properties in the loop. Will make this update.

client/src/lib/result.js
HaysCubing: Ok I cleaned this up and think I have it in a good spot. Just want to confirm a typo in your suggestion that what it should be is `const isClinched = currentResult.advancing && !currentResult.advancingQuestionable;`. Then the rest makes sense

client/src/lib/result.js
```diff
@@ -40,3 +47,84 @@ export function paddedAttemptResults(result, numberOfAttempts) {
...
+        toMonotonic(currentResult.best) === toMonotonic(prevResult.best)) {
+        // Rankings tie
+        currentResult.ranking = prevResult.ranking;
+      } else {
+        currentResult.ranking = i + 1;
+      }
+      const isClinched = currentResult.advancing && !currentResult.advancingQuestionable;
+      // A clinched result must still be clinched in the projected ranking,
+      // so we keep advancing state as is
+      if (!isClinched) {
```
HaysCubing: Add test for clinched scenarios before submitting

client/src/lib/result.js
```diff
@@ -40,3 +47,83 @@ export function paddedAttemptResults(result, numberOfAttempts) {
...
+        } else {
+          currentResult.advancing = false;
+          currentResult.advancingQuestionable = false;
+        }
+      }
+    }
+    else {
+      currentResult.advancing = false;
+      currentResult.advancingQuestionable = false;
+    }
```
jonatanklosko: If the result has no attempts, it's already `advancing = false`, `advancingQuestionable = false`, so we can keep it as is :)

```suggestion
```

Actually, I think the previous logic was correct:

```
if (currentResult.attempts.length === 0) {
  break;
}
```

Once we get to an empty result, it's all the rest is empty and their advancing state is unchanged.

client/src/lib/result.js
```diff
@@ -40,3 +47,83 @@ export function paddedAttemptResults(result, numberOfAttempts) {
...
+  });
+
+  // Sort based on projection with tiebreakers on single
+  resultsForView = orderBy(resultsForView, [
+    (result) => toMonotonic(result.projectedAverage),
+    (result) => toMonotonic(result.best),
+  ]);
+
+  if (isSkipped(resultsForView[0].projectedAverage)) {
+    // First place has no results. Do nothing.
```
jonatanklosko: Looking on the comment, we can make this more explicit:

```suggestion
  if (resultsForView[0].attempts.length === 0)) {
```

client/src/lib/result.js
```diff
@@ -40,3 +47,83 @@ export function paddedAttemptResults(result, numberOfAttempts) {
...
+}
+
+/**
+ * return the results object with additional properties:
+ * projectedAverage: projected final average based on current results
+ * forFirst: time needed to overtake first place
+ * forThird: time needed to overtake third place
+ */
+export function resultsForView(results, format, forecastView) {
+  console.log(results);
```
jonatanklosko:
```suggestion
```

client/src/lib/result.js
jonatanklosko: Gahh, yes I meant `const isClinched = currentResult.advancing && !currentResult.advancingQuestionable`. Sorry for the confusion :)

client/src/lib/result.js
jonatanklosko: Ah right, results with projected average have at least one attempt so they would be at the top above empty results. But yeah, doesn't hurt to have the sorting go first.

client/src/lib/attempt-result.js
```diff
@@ -124,6 +128,47 @@ function mean(values) {
...
+      return meanOfX(attemptResults);
+    }
+    if (attemptResults.length === 3) {
+      const [, x,] = attemptResults.slice().sort(compareAttemptResults);
+      return x;
+    }
+    if (attemptResults.length === 4) {
+      const [, x, y,] = attemptResults.slice().sort(compareAttemptResults);
+      return meanOfX([x, y]);
+    }
```
jonatanklosko, line 166: I would still do this for completeness, even if we don't use it directly:

```suggestion
    }
    return averageOf5(attemptResults);
```

client/src/lib/result.js
```diff
@@ -40,3 +47,83 @@ export function paddedAttemptResults(result, numberOfAttempts) {
...
+// Fewest moves is currently unsupported
+// Only final rounds are supported
+export function forecastViewEnabled(round) {
+  return round.format.sortBy != "best" &&
+    round.competitionEvent.event.id != "333fm" &&
+    round.advancementCondition === null;
+}
+
+export function resultProjectedAverage(result, format) {
+  if (isComplete(result.average)) {
```
jonatanklosko: It can also be DNF, so we want to check specifically for skipped:

```suggestion
  if (!isSkipped(result.average)) {
```

client/src/lib/attempt-result.js
jonatanklosko: We already have the average logic on the client anyway, so we want to make sure it's correct. I would let this function return the complete average also.

That said, using `result.average` if present as you did in `result.js` is a tiny optimisation, so it sounds good to me!

client/src/lib/tests/attempt-result.test.js
```diff
@@ -619,3 +620,26 @@ describe("incompleteMean", () => {
     expect(incompleteMean(attemptResults, "333fm")).toEqual(2200);
   });
 });
+
+describe("projectedAverage", () => {
+  var result;
```
jonatanklosko:
```suggestion
```

client/src/lib/tests/result.test.js
```diff
@@ -0,0 +1,58 @@
...
+            { attempts: [{ result: 100 }], best: 100, average: 0 },
+            { attempts: [{ result: 100 }], best: 99, average: 0  },
+            { attempts: [{ result: 102 }], best: 100, average: 0  },
+            { attempts: [{ result: 101 }], best: 100, average: 0  },
+            { attempts: [{ result: 101 }], best: 100, average: 0  }];
+            const expandedResults = resultsForView(results, format, true);
+            expect(expandedResults[0].projectedAverage).toEqual(100);
+            expect(expandedResults[0].best).toEqual(99);
+            expect(expandedResults[0].ranking).toEqual(1);
+            expect(expandedResults[0].advancingQuestionable).toEqual(true);
```
jonatanklosko: I believe you can do this:

```js
expect(expandedResults[0]).toMatchObject({
  projectedAverage: 100,
  best: 99,
  ranking: 1,
  advancingQuestionable: true,
});
```

Which makes the test easier to read :)

client/src/lib/result.js
HaysCubing: Ah right yeah you're correct

jonatanklosko (approved):
@HaysCubing thanks!

I updated toggle UI and did a few minor adjustments (mostly stylistic). I couldn't push to the branch (which you can allow/disallow when creating the PR), so I merged to main manually (9a46630b344a19da031cf4da769fbb3d63f40848) :)

PR #254, opened 2025-01-18
Change rounding to truncation for attempt results over 10 minutes

### LINKED ISSUE #253: Truncate, do not round, times >10 minutes and all 3x3x3 Multi-Blind times (Regulation 9f2)
Regulation [9f2 ](https://www.worldcubeassociation.org/regulations/full/#9f2)was changed to the following from January 1st, 2025:
_9f2) All timed results, averages, and means over 10 minutes, as well as all times for 3x3x3 Multi-Blind results, are measured and truncated to seconds._

Previously, the Regulation was:
_9f2) All timed results, averages, and means over 10 minutes, as well as all times for 3x3x3 Multi-Blind results, are measured and rounded to the nearest second (e.g. X.49 becomes X, X.50 becomes X+1)._

WCA Live currently implements the previous Regulation, and applies rounding in the specified circumstances. This is no longer appropriate following the Regulation change, and needs to be updated.

While knowledgeable users can bypass the issue by applying the necessary truncation before entering results into WCA Live, naïve users will not do so, and WCA Live will then record results incorrectly, not in accordance with the WCA Regulations. Therefore this a relatively important and urgent issue.

jonatanklosko:
Closes #253.

### DIFF
--- a/client/src/lib/attempt-result.js
+++ b/client/src/lib/attempt-result.js
@@ -92,9 +92,9 @@ export function average(attemptResults, eventId) {
 
   switch (attemptResults.length) {
     case 3:
-      return roundOver10Mins(meanOf3(attemptResults));
+      return truncateOver10Mins(meanOf3(attemptResults));
     case 5:
-      return roundOver10Mins(averageOf5(attemptResults));
+      return truncateOver10Mins(averageOf5(attemptResults));
     default:
       throw new Error(
         `Invalid number of attempt results, expected 3 or 5, got ${attemptResults.length}.`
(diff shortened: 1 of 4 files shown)


Nanush7:
@jonatanklosko I'm not sure if it's relevant, but aren't results being rounded here as well?

https://github.com/thewca/wca-live/blob/570f3beb7586f5b73efae5f0118abdd3a68fa4b8/client/src/lib/attempt-result.js#L214-L222

jonatanklosko:
@Nanush7 good catch, I fixed on main, thank you! I also searched for `round` to make sure there's no more hidden ones :) FWIW with this PR results given to `encodeMbldAttemptResult` should already have the centiseconds truncated (the form time field does that).

PR #256, opened 2025-02-07
Add "For First" and "For Third" columns to forecast view

HaysCubing:
Add 2 new statistics columns, "For First" and "For Third" to forecast view.

Columns show the time needed for a competitor to overtake 1st/3rd on their next solve.

All calculations are done using projected averages.

### DIFF
--- a/client/src/lib/attempt-result.js
+++ b/client/src/lib/attempt-result.js
@@ -4,8 +4,6 @@ import { shouldComputeAverage } from "./result";
 export const SKIPPED_VALUE = 0;
 export const DNF_VALUE = -1;
 export const DNS_VALUE = -2;
-export const NA_VALUE = -3;
-export const SUCCESS_VALUE = -4;
 
 export function isComplete(attemptResult) {
   return attemptResult > 0;
@@ -305,8 +303,6 @@ export function formatAttemptResult(attemptResult, eventId) {
   if (attemptResult === SKIPPED_VALUE) return "";
   if (attemptResult === DNF_VALUE) return "DNF";
   if (attemptResult === DNS_VALUE) return "DNS";
-  if (attemptResult === NA_VALUE) return "N/A";
-  if (attemptResult === SUCCESS_VALUE) return "SUCCESS";
   if (eventId === "333mbf") return formatMbldAttemptResult(attemptResult);
   if (eventId === "333fm") return formatFmAttemptResult(attemptResult);
   return centisecondsToClockFormat(attemptResult);
--- a/client/src/lib/attempt-result.js
+++ b/client/src/lib/attempt-result.js
@@ -4,6 +4,7 @@ import { shouldComputeAverage } from "./result";
 export const SKIPPED_VALUE = 0;
 export const DNF_VALUE = -1;
 export const DNS_VALUE = -2;
+export const NA_VALUE = -3;
 
 export function isComplete(attemptResult) {
   return attemptResult > 0;
@@ -123,9 +124,17 @@ function meanOfX(attemptResults) {
   return mean(attemptResults);
 }
 
+function sumOfX(attemptResults) {
+  if (!attemptResults.every(isComplete)) return DNF_VALUE;
+  return sum(attemptResults);
+}
+
 function mean(values) {
-  const sum = values.reduce((x, y) => x + y, 0);
-  return Math.round(sum / values.length);
+  return Math.round(sum(values) / values.length);
+}
+
+function sum(values) {
+  return values.reduce((x, y) => x + y, 0);
 }
 
 /**
@@ -165,6 +174,32 @@ export function projectedAverage(attemptResults, format) {
       return meanOfX([x, y]);
     }
     return averageOf5(attemptResults);
+  }
+
+  throw new Error("Unexpected format");
+}
+
+export function countingSum(attemptResults, format) {
+  if (attemptResults.length === 0) return Infinity;
+
+  if (format.numberOfAttempts === 3) {
+    return sumOfX(attemptResults);
+  }
+
+  if (format.numberOfAttempts === 5) {
+    if (attemptResults.length < 3) {
+      return sumOfX(attemptResults);
+    }
+    if (attemptResults.length === 3) {
+      const [, x] = attemptResults.slice().sort(compareAttemptResults);
+      return x;
+    }
+    if (attemptResults.length === 4) {
+      const [, x, y] = attemptResults.slice().sort(compareAttemptResults);
+      return sumOfX([x, y]);
+    }
+    const [, x, y, z] = attemptResults.slice().sort(compareAttemptResults);
+    return sumOfX([x, y, z]);
   }
 
   throw new Error("Unexpected format");
@@ -303,6 +338,7 @@ export function formatAttemptResult(attemptResult, eventId) {
   if (attemptResult === SKIPPED_VALUE) return "";
   if (attemptResult === DNF_VALUE) return "DNF";
   if (attemptResult === DNS_VALUE) return "DNS";
+  if (attemptResult === NA_VALUE) return "N/A";
   if (eventId === "333mbf") return formatMbldAttemptResult(attemptResult);
   if (eventId === "333fm") return formatFmAttemptResult(attemptResult);
   return centisecondsToClockFormat(attemptResult);
--- a/client/src/lib/attempt-result.js
+++ b/client/src/lib/attempt-result.js
@@ -179,8 +179,9 @@ export function projectedAverage(attemptResults, format) {
   throw new Error("Unexpected format");
 }
 
+// Needs to assume inputs are all valid (remove function sumOfX)
 export function countingSum(attemptResults, format) {
-  if (attemptResults.length === 0) return Infinity;
+  if (attemptResults.length === 0) return 0;
 
   if (format.numberOfAttempts === 3) {
     return sumOfX(attemptResults);
--- a/client/src/lib/attempt-result.js
+++ b/client/src/lib/attempt-result.js
@@ -124,16 +124,11 @@ function meanOfX(attemptResults) {
   return mean(attemptResults);
 }
 
-function sumOfX(attemptResults) {
-  if (!attemptResults.every(isComplete)) return DNF_VALUE;
-  return sum(attemptResults);
-}
-
 function mean(values) {
   return Math.round(sum(values) / values.length);
 }
 
-function sum(values) {
+export function sum(values) {
   return values.reduce((x, y) => x + y, 0);
 }
 
@@ -174,33 +169,6 @@ export function projectedAverage(attemptResults, format) {
       return meanOfX([x, y]);
     }
     return averageOf5(attemptResults);
-  }
-
-  throw new Error("Unexpected format");
-}
-
-// Needs to assume inputs are all valid (remove function sumOfX)
-export function countingSum(attemptResults, format) {
-  if (attemptResults.length === 0) return 0;
-
-  if (format.numberOfAttempts === 3) {
-    return sumOfX(attemptResults);
-  }
-
-  if (format.numberOfAttempts === 5) {
-    if (attemptResults.length < 3) {
-      return sumOfX(attemptResults);
-    }
-    if (attemptResults.length === 3) {
-      const [, x] = attemptResults.slice().sort(compareAttemptResults);
-      return x;
-    }
-    if (attemptResults.length === 4) {
-      const [, x, y] = attemptResults.slice().sort(compareAttemptResults);
-      return sumOfX([x, y]);
-    }
-    const [, x, y, z] = attemptResults.slice().sort(compareAttemptResults);
-    return sumOfX([x, y, z]);
   }
 
   throw new Error("Unexpected format");
--- a/client/src/lib/attempt-result.js
+++ b/client/src/lib/attempt-result.js
@@ -5,6 +5,7 @@ export const SKIPPED_VALUE = 0;
 export const DNF_VALUE = -1;
 export const DNS_VALUE = -2;
 export const NA_VALUE = -3;
+export const SUCCESS_VALUE = -4;
 
 export function isComplete(attemptResult) {
   return attemptResult > 0;
@@ -18,7 +19,7 @@ export function toMonotonic(attemptResult) {
   return isComplete(attemptResult) ? attemptResult : Infinity;
 }
 
-function compareAttemptResults(attemptResult1, attemptResult2) {
+export function compareAttemptResults(attemptResult1, attemptResult2) {
   if (!isComplete(attemptResult1) && !isComplete(attemptResult2)) return 0;
   if (!isComplete(attemptResult1) && isComplete(attemptResult2)) return 1;
   if (isComplete(attemptResult1) && !isComplete(attemptResult2)) return -1;
@@ -125,11 +126,8 @@ function meanOfX(attemptResults) {
 }
 
 function mean(values) {
-  return Math.round(sum(values) / values.length);
-}
-
-export function sum(values) {
-  return values.reduce((x, y) => x + y, 0);
+  const sum = values.reduce((x, y) => x + y, 0);
+  return Math.round(sum / values.length);
 }
 
 /**
@@ -308,6 +306,7 @@ export function formatAttemptResult(attemptResult, eventId) {
   if (attemptResult === DNF_VALUE) return "DNF";
   if (attemptResult === DNS_VALUE) return "DNS";
   if (attemptResult === NA_VALUE) return "N/A";
+  if (attemptResult === SUCCESS_VALUE) return "SUCCESS";
   if (eventId === "333mbf") return formatMbldAttemptResult(attemptResult);
   if (eventId === "333fm") return formatFmAttemptResult(attemptResult);
   return centisecondsToClockFormat(attemptResult);
--- a/client/src/lib/result.js
+++ b/client/src/lib/result.js
@@ -86,7 +86,7 @@ export function forecastViewSupported(round) {
  * Wrapper for formatAttemptResult. Also handles values
  * NA_VALUE and SUCCESS_VALUE
  */
-export function formatAttemptResultForN(attemptResult, eventId) {
+export function formatAttemptResultForView(attemptResult, eventId) {
   if (attemptResult === NA_VALUE) return "N/A";
   if (attemptResult === SUCCESS_VALUE) return "SUCCESS";
   return formatAttemptResult(attemptResult, eventId);
--- a/client/src/lib/result.js
+++ b/client/src/lib/result.js
@@ -218,10 +218,8 @@ export function timeNeededToOvertake(result, format, overtakeResult) {
 
   let attemptResults = result.attempts.map((attempt) => attempt.result);
   const resultWorst = attemptResults.slice().sort(compareAttemptResults).pop();
-  const bestComparison = compareAttemptResults(
-    result.best,
-    overtakeResult.best
-  );
+  const betterBest =
+    compareAttemptResults(result.best, overtakeResult.best) < 0;
 
   // Projection will change from a mean to a median after a time is added
   if (attemptResults.length === 2 && format.numberOfAttempts === 5) {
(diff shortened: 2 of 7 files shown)


HaysCubing:
@jonatanklosko let me know what you think of this change. It's more drastic than the first "forecast view" change since it adds two new statics columns. I think this will be really nice to have for the projector display at worlds so people in the crowd don't have to do the math on what so and so needs to win / podium.

It should be pretty clean already, but wanted to get general ideas and comments out of the way now before I write tests and add documentation

client/src/lib/result.js
```diff
@@ -135,9 +153,93 @@ export function resultsForView(results, format, forecastView) {
...
+          );
+        }
+      }
+    }
+  }
+
   return resultsForView;
 }
 
+export function timeNeededToOvertake(
```
HaysCubing: Wasn't sure if parts of this should be functioned out to attempt-result.js

client/src/lib/result.js
```diff
@@ -1,32 +1,48 @@
...
+  DNF_VALUE,
 } from "./attempt-result";
 
 /**
  * Returns a list of objects corresponding to result statistics - best and average.
  * The first statistic is the one that determines the ranking.
  * This is a common logic used in all result tables/dialogs.
  */
-export function orderedResultStats(eventId, format) {
+export function orderedResultStats(eventId, format, forecastView = null) {
```
jonatanklosko:
```suggestion
export function orderedResultStats(eventId, format, forecastView = false) {
```

client/src/lib/result.js
```diff
@@ -1,32 +1,48 @@
...
   const { numberOfAttempts, sortBy } = format;
 
   if (!shouldComputeAverage(eventId, numberOfAttempts)) {
     return [{ name: "Best", field: "best", recordTagField: "singleRecordTag" }];
   }
 
-  const stats = [
+  var stats = (stats = [
```
jonatanklosko: When defining variables, prefer to use `const` and `let` (when you reassign the name later). `var` is an old specifier and has its issues, which was the reason for introducing `const` and `let` :)

```suggestion
  let stats = [
```

(same for all other `var` usages)

client/src/lib/attempt-result.js
```diff
@@ -124,8 +125,11 @@ function meanOfX(attemptResults) {
 }
 
 function mean(values) {
-  const sum = values.reduce((x, y) => x + y, 0);
-  return Math.round(sum / values.length);
+  return Math.round(sum(values) / values.length);
+}
+
+export function sum(values) {
+  return values.reduce((x, y) => x + y, 0);
```
jonatanklosko: It's a detail, but I wouldn't export this function from this module, because it's not specific to attempt results at all. We could extract to utils, but I don't think it's worth it either, I would just redefine the function in the other module, just to avoid writing the `reduce` inline.

client/src/lib/result.js
```diff
@@ -135,9 +153,93 @@ export function resultsForView(results, format, forecastView) {
     prevResult = currentResult;
   }
 
+  const secondComplete =
+    resultsForView.length > 1 && isComplete(resultsForView[1].projectedAverage);
+  const fourthComplete =
+    resultsForView.length > 3 && isComplete(resultsForView[3].projectedAverage);
+  if (secondComplete) {
```
jonatanklosko: I am missing why we check for second and fourth completeness. Could you explain?

Also note that 4th can be complete, even if 2nd is note, because we rerank based on projected average.

client/src/lib/result.js
```diff
@@ -135,9 +153,93 @@ export function resultsForView(results, format, forecastView) {
...
+    for (let i = 0; i < resultsForView.length; i++) {
+      var result = resultsForView[i];
+      if (isSkipped(result.projectedAverage)) {
+        break;
+      }
```
jonatanklosko: I would make the same check we do in the prior loop, which is equivalent.

```suggestion
      if (currentResult.attempts.length === 0) {
        // From this point forward all results are empty, so we are done
        break;
      }
```

client/src/lib/result.js
```diff
@@ -135,9 +153,93 @@ export function resultsForView(results, format, forecastView) {
...
+      if (isSkipped(result.average)) {
+        var firstIndex = i == 0 ? 1 : 0;
```
jonatanklosko: If this result is the first, I would probably just set `forFirst = 0`, because it just doesn't apply.

If we wanted to show DNF, I would perhaps do this:

```elixir
result.forFirst = timeNeededToOvertake(
  result,
  format,
  resultsForView[0].projectedAverage,
  resultsForView[0].best
);

if (i === 0 && !isSkipped(result.forFirst)) {
  result.forFirst = DNF_VALUE;
}
```

So if it's the first result, we use `timeNeededToOvertake` effectively to check if the stat is computed at all, in which case we set DNF, because we know we are already first.

But again, I think setting 0 should be totally fine.

(same for `forThird`)

client/src/lib/result.js
```diff
@@ -135,9 +153,93 @@ export function resultsForView(results, format, forecastView) {
...
+  }
+
   return resultsForView;
 }
 
+export function timeNeededToOvertake(
+  result,
+  format,
+  overtakeAverage,
+  overtakeBest
```
jonatanklosko: Since you accept `result` as the argument, let's also accept `overtakeResult` for consistency (instead of passing average and best separately).

client/src/lib/result.js
```diff
@@ -135,9 +153,93 @@ export function resultsForView(results, format, forecastView) {
...
+
+  const isMean = format.numberOfAttempts === 3 || result.attempts.length < 2;
+  const nextCountingSolves = result.attempts.length + (isMean ? 1 : -1);
+  const totalNeeded = overtakeAverage * nextCountingSolves;
+  // For a mean of 3, .01 can be added to achieve the same rounded result
+  const roundingBuffer = nextCountingSolves === 3 ? 1 : 0;
+  var countingSum = sum(attemptResults);
+  if (!isMean) {
+    countingSum =
+      countingSum - Math.min(...attemptResults) - Math.max(...attemptResults);
```
jonatanklosko: Min/max here needs to account for DNF (-1).

client/src/lib/result.js
jonatanklosko: Given that we deal with both best and average, I think it belongs here just fine.

Btw. you don't necessarily need to export this function, you could write tests for `resultsForView` and this function is just an implementation detail. But if testing this function makes it easier to cover edge cases, then that's fine as well.

jonatanklosko (commented):
Hey @HaysCubing, thanks for working on this, here's a few comments.

We definitely need tests :D

client/src/lib/result.js
HaysCubing: Ah didn't know that about var, let, and const. Thanks!

client/src/lib/result.js
HaysCubing: You have to export functions to use them in tests in js? Interesting.

I didn't export for now - I'll reevaluate when I write tests

client/src/lib/result.js
HaysCubing: Since we rerank on projected average, doesn't that mean if 4th is complete 2nd has to be complete as well? All of the completed projected averages will be listed first.

client/src/lib/result.js
HaysCubing: So the idea here was that if the first/third positions aren't complete, we still want to see what time they need to stay in their current position (since their average could get worse on the next solve). I think this is pretty important to have since it covers the case where the projected winner is on their last solve and we still want to know what they need to secure the win.

I'm realizing it's going to make the logic a bit annoying here because 1st/3rd get compared with 2nd/4th.

client/src/lib/result.js
HaysCubing: Yeah wow that's way cleaner

client/src/lib/result.js
HaysCubing: Cleaned this up a lot. Now function `timeNeededToOvertake` can handle incomplete values and so we don't need to check if certain values are complete or not. Loop just needs to make sure it won't hit an out of bounds error.

HaysCubing:
Still working on this iteration. `timeNeededToOvertake` is getting complicated since now it handles incomplete results and will definitely need to have rigorous tests.

Looks like you're on board with the general ideas though so I'll go ahead and write tests and document everything for my next iteration ready for review. Will have time later in the next week.

client/src/lib/result.js
HaysCubing: Actually not as complicated after I updated `timeNeededToOvertake` to handle skipped and incomplete values

client/src/lib/result.js
jonatanklosko: Ah, the check was on projectedAverage, not average, I get it.

client/src/lib/attempt-result.js
```diff
@@ -303,6 +305,8 @@ export function formatAttemptResult(attemptResult, eventId) {
   if (attemptResult === SKIPPED_VALUE) return "";
   if (attemptResult === DNF_VALUE) return "DNF";
   if (attemptResult === DNS_VALUE) return "DNS";
+  if (attemptResult === NA_VALUE) return "N/A";
+  if (attemptResult === SUCCESS_VALUE) return "SUCCESS";
```
jonatanklosko: I would prefer to only use the standardized values in this module (i.e. we don't expect -3 or -4 to ever be stored as an actual value in the database). I was on the fence about `NA_VALUE`, but now that there is `SUCCESS_VALUE` I think we should definitely do this elsewhere.

I think what I would do instead, is to have the `NA_VALUE` and `SUCCESS_VALUE` constants in `result.js`, and in that same file define `formatAttemptResultForN`. It would check for `NA_VALUE`/`SUCCESS_VALUE` first, and fallback to `formatAttemptResult` otherwise. In `ResultStat` component we can special case the `forFirst`/`forThird` fields to use `formatAttemptResultForN`.

Does that make sense?

client/vite.config.js
```diff
@@ -17,6 +17,6 @@ export default defineConfig({
     },
     // Uncomment when running using Docker on Windows.
     // See https://github.com/vitejs/vite/issues/1153#issuecomment-785467271
-    // watch: { usePolling: true },
+    watch: { usePolling: true },
```
HaysCubing: Undo this change before submitting

client/src/lib/attempt-result.js
HaysCubing: Yep sounds good!

I added function `formatAttemptResultForN` and just changed all references in `ResultStat.jsx` to use it. Since it's just a wrapper around `formatAttemptResult` I figured it would be cleaner and functionally the same but let me know if I should add special logic in `ResultStat.jsx` to call the different versions.

client/src/lib/tests/result.test.js
```diff
@@ -148,4 +188,341 @@ describe("viewResults", () => {
...
+});
+
+describe("timeNeededToOvertake", () => {
+  it("returns DNF if overtake result is skipped", () => {
+    expect(timeNeededToOvertake(null, null, { projectedAverage: 0 })).toEqual(
+      -1
+    );
+  });
+
+  it("handles results turning from a mean to an average", () => {
```
jonatanklosko:
```suggestion
  it("handles projected average turning from a mean to an average", () => {
```

client/src/lib/tests/result.test.js
```diff
@@ -148,4 +188,341 @@ describe("viewResults", () => {
...
+          best: 110,
+          projectedAverage: 110,
+          attempts: [{ result: 110 }, { result: 110 }],
+        },
+        format,
+        overtakeResult
+      )
+    ).toEqual(81);
+    overtakeResult = { best: 50, projectedAverage: 100 };
+    expect(
```
jonatanklosko, line 425: It would be helpful to have a short comment before each case outlining the case, for example "Can overtake with better mean and worse single", "Can overtake with same mean, and getting a better single", etc. I would also add an extra newline before the cases.

Those could be separate tests altogether, but it's also fine as is, it's just about making it easier for the reader to understand which scenario are we asserting (instead of reverse-figuring it out by analyzing the times).

client/src/lib/result.js
```diff
@@ -63,13 +82,29 @@ export function forecastViewSupported(round) {
   );
 }
 
+/**
+ * Wrapper for formatAttemptResult. Also handles values
+ * NA_VALUE and SUCCESS_VALUE
+ */
+export function formatAttemptResultForN(attemptResult, eventId) {
```
jonatanklosko: Since we are using it for in all cases, I would call it `formatAttemptResultForView`, which makes it clear it's a generic version.

jonatanklosko (commented):
The logic looks correct to me, just a few minor comments and I think it's good to go :)

client/src/lib/tests/result.test.js
HaysCubing: Yeah I was thinking the same - wasn't sure what the proper style was for these tests.

Added comments

HaysCubing: Adding newline seems unnecessary when there is a comment already - let me know if you want me to add that still

client/src/lib/tests/result.test.js
```diff
@@ -148,4 +188,366 @@ describe("viewResults", () => {
...
+describe("timeNeededToOvertake", () => {
+  it("returns DNF if overtake result is skipped", () => {
+    expect(timeNeededToOvertake(null, null, { projectedAverage: 0 })).toEqual(
+      -1
+    );
+  });
+
+  it("handles projected average turning from a mean to an average", () => {
+    const format = { numberOfAttempts: 5 };
+    let result = { attempts: [{ result: 99 }, { result: 100 }], best: 99 };
```
jonatanklosko, line 259:
```suggestion
    const format = { numberOfAttempts: 5 };
    let result = { attempts: [{ result: 99 }, { result: 100 }], best: 99 };

```

client/src/lib/tests/result.test.js
```diff
@@ -148,4 +188,366 @@ describe("viewResults", () => {
...
+    // Worst possible average beats overtake average
+    expect(
+      timeNeededToOvertake(result, format, { best: 10, projectedAverage: 101 })
+    ).toEqual(-1);
+    // Worst possible average ties overtake average, wins on best
+    expect(
+      timeNeededToOvertake(result, format, { best: 100, projectedAverage: 100 })
+    ).toEqual(-1);
+    result = { attempts: [{ result: 100 }, { result: -1 }], best: 100 };
```
jonatanklosko, line 270:
```suggestion

    result = { attempts: [{ result: 100 }, { result: -1 }], best: 100 };

```

client/src/lib/tests/result.test.js
```diff
@@ -148,4 +188,366 @@ describe("viewResults", () => {
...
+    ).toEqual(-4);
+    // Best possible average ties overtake average
+    expect(
+      timeNeededToOvertake(result, format, { best: 50, projectedAverage: 100 })
+    ).toEqual(49);
+    // Best possible average loses to overtake average
+    expect(
+      timeNeededToOvertake(result, format, { best: 50, projectedAverage: 50 })
+    ).toEqual(-3);
+    result = { attempts: [{ result: -1 }, { result: -1 }], best: -1 };
```
jonatanklosko, line 293:
```suggestion

    result = { attempts: [{ result: -1 }, { result: -1 }], best: -1 };

```

client/src/lib/tests/result.test.js
```diff
@@ -148,4 +188,366 @@ describe("viewResults", () => {
...
+    // Best possible average ties overtake average, overtake average incomplete
+    expect(
+      timeNeededToOvertake(result, format, { best: -1, projectedAverage: -1 })
+    ).toEqual(-4);
+  });
+
+  it("handles incomplete overtake results", () => {
+    let overtakeResult = { best: 50, projectedAverage: -1 };
+    const format = { numberOfAttempts: 5 };
```
jonatanklosko, line 303:
```suggestion
    const format = { numberOfAttempts: 5 };

```

client/src/lib/tests/result.test.js
```diff
@@ -148,4 +188,366 @@ describe("viewResults", () => {
...
+        {
+          best: 100,
+          projectedAverage: 100,
+          attempts: [{ result: 100 }],
+        },
+        format,
+        overtakeResult
+      )
+    ).toEqual(-4);
+    overtakeResult = { best: -1, projectedAverage: -1 };
```
jonatanklosko, line 360:
```suggestion

    overtakeResult = { best: -1, projectedAverage: -1 };

```

client/src/lib/tests/result.test.js
```diff
@@ -148,4 +188,366 @@ describe("viewResults", () => {
...
+        },
+        format,
+        overtakeResult
+      )
+    ).toEqual(-4);
+  });
+
+  it("handles incomplete results", () => {
+    const overtakeResult = { best: 50, projectedAverage: 50 };
+    const format = { numberOfAttempts: 3 };
```
jonatanklosko, line 378:
```suggestion
    const format = { numberOfAttempts: 3 };

```

client/src/lib/tests/result.test.js
```diff
@@ -148,4 +188,366 @@ describe("viewResults", () => {
...
+        },
+        format,
+        overtakeResult
+      )
+    ).toEqual(-3);
+  });
+
+  it("handles overtake for mean", () => {
+    let overtakeResult = { best: 100, projectedAverage: 100 };
+    let format = { numberOfAttempts: 3 };
```
jonatanklosko, line 395:
```suggestion
    let format = { numberOfAttempts: 3 };

```

client/src/lib/tests/result.test.js
```diff
@@ -148,4 +188,366 @@ describe("viewResults", () => {
...
+        {
+          best: 110,
+          projectedAverage: 110,
+          attempts: [{ result: 110 }, { result: 110 }],
+        },
+        format,
+        overtakeResult
+      )
+    ).toEqual(81);
+    overtakeResult = { best: 50, projectedAverage: 100 };
```
jonatanklosko, line 422:
```suggestion

    overtakeResult = { best: 50, projectedAverage: 100 };

```

client/src/lib/tests/result.test.js
```diff
@@ -148,4 +188,366 @@ describe("viewResults", () => {
...
+          best: 110,
+          projectedAverage: 110,
+          attempts: [{ result: 110 }, { result: 110 }],
+        },
+        format,
+        overtakeResult
+      )
+    ).toEqual(78);
+
+    overtakeResult = { best: 80, projectedAverage: 100 };
```
jonatanklosko, line 449:
```suggestion
    overtakeResult = { best: 80, projectedAverage: 100 };

```

client/src/lib/tests/result.test.js
```diff
@@ -148,4 +188,366 @@ describe("viewResults", () => {
...
+          projectedAverage: 110,
+          attempts: [{ result: 110 }, { result: 110 }],
+        },
+        format,
+        overtakeResult
+      )
+    ).toEqual(79);
+
+    format = { numberOfAttempts: 5 };
+    overtakeResult = { best: 100, projectedAverage: 100 };
```
jonatanklosko, line 465:
```suggestion
    overtakeResult = { best: 100, projectedAverage: 100 };

```

client/src/lib/tests/result.test.js
```diff
@@ -148,4 +188,366 @@ describe("viewResults", () => {
...
+        },
+        format,
+        overtakeResult
+      )
+    ).toEqual(90);
+  });
+
+  it("handles overtake for average", () => {
+    let overtakeResult = { best: 50, projectedAverage: 100 };
+    const format = { numberOfAttempts: 5 };
```
jonatanklosko, line 483:
```suggestion
    const format = { numberOfAttempts: 5 };

```

jonatanklosko (commented):
I would add newlines around the reassignments. I started reading and got confused about an `expect`, and only then realised the result is different. It is easy to miss with no spacing.

jonatanklosko (approved):
Thank you :)

PR #259, opened 2025-03-07
Update forecast view to support 333fm

HaysCubing:
Add fewest moves support to forecast view. Just add support for ranking by projected average. "For First" and "For Third" are not as relevant since every solve is concurrent anyways. Also I don't feel like supporting that logic. I think the most helpful thing is to just rank on projected average.


jonatanklosko (approved):
Looks good, thanks!

PR #261, opened 2025-03-07
Support forecast view for non final rounds with "ranking " advancement condition type

HaysCubing:
Specifically adding this functionality for 3x3 semifinals at worlds. This will be treated similar to the other finals and would like to be able to see what people need to get in order to jump into the top 16.

Coded in a way that is generic for all rounds with a "ranking" advancement condition type

### DIFF
--- a/client/src/lib/result.js
+++ b/client/src/lib/result.js
@@ -42,13 +42,13 @@ export function orderedResultStats(
   stats = sortBy === "best" ? stats : stats.reverse();
   if (forecastView && eventId != "333fm") {
     stats.push({
-      name: "For 1st",
+      name: "For 1",
       field: "forFirst",
     });
     stats.push({
       name: advancementCondition
-        ? "Adv. [" + advancementCondition.level + "]"
-        : "For 3rd",
+        ? "For " + advancementCondition.level
+        : "For 3",
       field: "forAdvance",
     });
   }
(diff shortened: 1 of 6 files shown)


client/src/lib/result.js
```diff
@@ -41,8 +46,10 @@ export function orderedResultStats(eventId, format, forecastView = false) {
       field: "forFirst",
     });
     stats.push({
-      name: "For 3rd",
-      field: "forThird",
+      name: advancementCondition
```
HaysCubing, line 49: Wasn't sure the best way to format this. Feel free to change to whatever. Current ex: "Adv. [16]"

jonatanklosko: Perhaps we could make them "For 1.", "For 3." and "For {level}."? It's fine without the dot also, since we don't include it in the ranking column either.

client/src/lib/result.js
```diff
@@ -172,6 +183,8 @@ export function resultsForView(results, eventId, format, forecastView) {
     prevResult = currentResult;
   }
 
+  // Default to podium (top 3) if no advancement condition
+  let advancementIndex = advancementCondition?.level ?? 3;
```
jonatanklosko: Let's call this `advancementRanking`, because it's 1-based position.

jonatanklosko (approved):
Thanks!

PR #263, opened 2025-03-16
Change order of stats columns in forecast view

HaysCubing:
BPA/WPA are only valid when 4/5 solves are complete, while For1/ForX are always valid. Change For1/ForX to go first


jonatanklosko (approved):
Sounds good, thanks!

PR #271, opened 2025-05-22
Show only advancing results in projector with forecast view

HaysCubing:
Forecast view is specifically for showing projected results in a real time environment. 

For projector view, linger on each page for longer, populate the page faster, and only show results that could potentially advance.

### DIFF
--- a/client/src/components/ResultsProjector/ResultsProjector.jsx
+++ b/client/src/components/ResultsProjector/ResultsProjector.jsx
@@ -134,17 +134,18 @@ function ResultsProjector({
         setStatus(STATUS.SHOWING);
         setTopResultIndex((topResultIndex) => {
           const newIndex = topResultIndex + getNumberOfRows();
-          const maxIndex =
-            forecastView && advancementCondition
-              ? advancementCondition.level * 2
-              : nonemptyResults.length;
-          return newIndex >= maxIndex ? 0 : newIndex;
+          return newIndex > nonemptyResults.length ||
+            (forecastView &&
+              !nonemptyResults[topResultIndex].advancing &&
+              !nonemptyResults[newIndex].advancing)
+            ? o
+            : newIndex;
         });
       }, DURATION.HIDING);
       return () => clearTimeout(timeout);
     }
     throw new Error(`Unrecognized status: ${status}`);
-  }, [status, nonemptyResults.length]);
+  }, [status, nonemptyResults, forecastView]);
 
   return (
     <Dialog
--- a/client/src/components/ResultsProjector/ResultsProjector.jsx
+++ b/client/src/components/ResultsProjector/ResultsProjector.jsx
@@ -70,6 +70,7 @@ const STATUS = {
 
 const DURATION = {
   SHOWN: 10 * 1000,
+  FORECAST_SHOWN: 20 * 1000,
   SHOWING: 1000,
   HIDING: 1000,
 };
@@ -111,9 +112,12 @@ function ResultsProjector({
     }
     if (status === STATUS.SHOWN) {
       if (nonemptyResults.length > getNumberOfRows()) {
-        const timeout = setTimeout(() => {
-          setStatus(STATUS.HIDING);
-        }, DURATION.SHOWN);
+        const timeout = setTimeout(
+          () => {
+            setStatus(STATUS.HIDING);
+          },
+          forecastView ? DURATION.FORECAST_SHOWN : DURATION.SHOWN
+        );
         return () => clearTimeout(timeout);
       } else {
         return;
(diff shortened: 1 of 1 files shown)


client/src/components/ResultsProjector/ResultsProjector.jsx
```diff
@@ -130,7 +134,11 @@ function ResultsProjector({
         setStatus(STATUS.SHOWING);
         setTopResultIndex((topResultIndex) => {
           const newIndex = topResultIndex + getNumberOfRows();
-          return newIndex >= nonemptyResults.length ? 0 : newIndex;
+          const maxIndex =
+            forecastView && advancementCondition
+              ? advancementCondition.level * 2
```
jonatanklosko: I may be missing the point, but if 12 people advance, showing top 24 feels very arbitrary to me. Also, I would prefer not to assume the advancement condition is "ranking" here.

What would make more sense is to check if `nonemptyResults[newIndex].advancing` is true (regardless if questionable or not), so once we page into results that definitely don't advance, we go back to page 1.

That said, I am not convinced we should dictate such behaviour. Perhaps it could be optional clicking an extra icon specifically in the projector view top bar, where the option would be to only show advancing results (similarly to how forecast view is toggled).

HaysCubing: Agree on not assuming advancement condition is "ranking". It's not clear here that advancementCondition.level is only safe because it's gated behind forecastView.

Also yeah I arbitrarily chose 2*advancing to show.

So the point here is that at worlds we are planning on doing some sort of semi-live data entry for semifinals. During this time I want to have WCA live being projected for the audience in the venue to see. Since 20 people will be advancing to finals, I want to show all results that are in the running for finals, but not the others (ex. showing ranks 150+ during 2x2 semifinals is pretty useless).

In order to do this I need some sort of scrolling capability but only for relevant results. What do you think the best way to handle this is? I agree that just chucking this as a default for forecastView is not the best style and it would probably make more sense as a separate feature. However since forecastView is pretty experimental and [forecastView + projector + scroll] is very specific to this type of situation I figured it would also be ok just to lump this all together.

To make the scrolling requirements less arbitrary, I was thinking it would show a single page of results where the `nonemptyResults[newIndex].advancing` is false, and then cycle back to the start.

HaysCubing: Code updated where it cycles if the current page is the second page starting with a non advancement.

client/src/components/ResultsProjector/ResultsProjector.jsx
```diff
@@ -130,13 +134,18 @@ function ResultsProjector({
...
+              !nonemptyResults[newIndex].advancing)
+            ? 0
+            : newIndex;
         });
       }, DURATION.HIDING);
       return () => clearTimeout(timeout);
     }
     throw new Error(`Unrecognized status: ${status}`);
-  }, [status, nonemptyResults.length]);
+  }, [status, nonemptyResults, forecastView]);
```
jonatanklosko: This is the list of values that the `useEffect` depends on, whenever it changes it will run the cleanup (in this case the `clearTimeout`s) and re-execute with new values. Putting `nonemptyResults` as a dependency, means the timeouts will reset on every time entered, which is not what we want.

This is a bit annoying, but what I would do in this case is this:

```js
const nonemptyResultsRef = useRef(nonemptyResults);

useEffect(() => {
  nonemptyResultsRef.current = nonemptyResults;
});

useEffect(() => {
  const nonemptyResults = nonemptyResultsRef.current;

  // the current code...
}, [status, forecastView])
```

This way the effect doesn't **depend** on `nonemptyResults` (thus doesn't reset when it changes), but on every status change it uses the latest value of `nonemptyResults`.

client/src/components/ResultsProjector/ResultsProjector.jsx
```diff
@@ -130,13 +134,18 @@ function ResultsProjector({
         setStatus(STATUS.SHOWING);
         setTopResultIndex((topResultIndex) => {
           const newIndex = topResultIndex + getNumberOfRows();
-          return newIndex >= nonemptyResults.length ? 0 : newIndex;
+          return newIndex > nonemptyResults.length ||
+            (forecastView &&
```
jonatanklosko, line 147: Let's document the behaviour with a comment:

```suggestion
            // When forecast view is enabled, the focus is usually on the advancing
            // results, so we only show a single page of non-advancing results and
            // roll back to the first page afterwards.
            (forecastView &&
```

jonatanklosko (commented):
Sorry for the late reply, two more comments :)

client/src/components/ResultsProjector/ResultsProjector.jsx
HaysCubing: Went with a slightly different solution as `useRef` wasn't recognized. Also just used a `let` for the ref variable since it is being changed constantly.

HaysCubing: Never mind the compiler didn't like my solution. I just imported `useRef` and what you had works.

jonatanklosko (approved):

PR #279, opened 2025-07-08
Enhance mobile UI/UX

coder13:
Mobile devices are one of the most useful devices to use wca live on, but the experience has flaws.

Among them, this PR addresses some:

- table makes the entire result entry page scroll horizontally
  - This makes the double check button harder to access. This would be the most useful page for a user on mobile to visit
- double check page: center contents horizontally
  - I personally feel this makes the page present a little bit nicer on mobile
- double check page: swipe in between results 
  - This makes the UI feel a bit more natural bringing natural mobile gestures to a mobile view

image

image


jonatanklosko:
Hey @coder13, thanks for the improvements!

I think some of the width/height changes are not necessary, I simplified it a bit. It the vertical scroll and admin table looks good from what I checked, if there is a scenario I am missing let me know!

jonatanklosko:
I'm going to merge this, if I missed anything let me know :)

PR #280, opened 2025-07-08
Persist batch mode setting

### LINKED ISSUE #273: Persist batch mode in localStorage
## Issue

When entering scorecards, it is often necessary to refresh the page (e.g. to get new data without accidentally overwriting someone else's work).  Currently this causes batch mode to revert to the off state (assuming you had no in-progress batch, which is usually the case when refreshing).  At a venue with poor internet (and/or high data needs), this causes extra frustration when accidentally entering the first scorecard after refreshing.

I had a few, not-mutually-exclusive ideas to fix this:
1. persist batch mode choice on refresh
2. add a way to refresh data w/o refreshing page (`meta`+`.` maybe?)
3. add an obvious style change while in batch mode - eg highlight background of the `ResultAttemptsForm`

## Proposed Code for Solution 1

Here's the code I wrote for solution 1.  I don't have permissions to create a branch, otherwise I'd just open a PR.  Let me know if there's a better way to do this!

```diff
diff --git a/client/src/components/admin/AdminRound/AdminRoundContent.jsx b/client/src/components/admin/AdminRound/AdminRoundContent.jsx
index 84db5a8..3b39176 100644
--- a/client/src/components/admin/AdminRound/AdminRoundContent.jsx
+++ b/client/src/components/admin/AdminRound/AdminRoundContent.jsx
@@ -40,8 +40,23 @@ const ENTER_RESULTS = gql`
   ${ADMIN_ROUND_RESULT_FRAGMENT}
 `;
 
-// Persist batch results in local storage, in case people navigate
-// or refresh the page
+const IS_BATCH_MODE_KEY = `wca-live:is-batch-mode`;
+/**
+ * Persist whether batch mode is toggled on, in case people navigate
+ * or refresh the page
+ */
+function getStoreIsBatchMode() {
+  return localStorage.getItem(IS_BATCH_MODE_KEY);
+}
+
+function setStoreIsBatchMode(isBatchMode) {
+  localStorage.setItem(IS_BATCH_MODE_KEY, isBatchMode)
+}
+
+/**
+ * Persist batch results in local storage, in case people navigate
+ * or refresh the page
+ */
 
 function getStoreBatchResults(roundId) {
   const json = localStorage.getItem(`wca-live:batch-results:${roundId}`);
@@ -71,7 +86,7 @@ function AdminRoundContent({ round, competitionId, officialWorldRecords }) {
   const [batchResults, setBatchResults] = useState(() =>
     getStoreBatchResults(round.id),
   );
-  const [isBatchMode, setIsBatchMode] = useState(batchResults.length > 0);
+  const [isBatchMode, setIsBatchMode] = useState(batchResults.length > 0 || getStoreIsBatchMode());
   const formContainerRef = useRef(null);
 
   const [enterResults, { loading }] = useMutation(ENTER_RESULTS, {
@@ -189,7 +204,10 @@ function AdminRoundContent({ round, competitionId, officialWorldRecords }) {
               control={
                 <Checkbox
                   checked={isBatchMode}
-                  onChange={(event) => setIsBatchMode(event.target.checked)}
+                  onChange={(event) => {
+                    setIsBatchMode(event.target.checked);
+                    setStoreIsBatchMode(event.target.checked);
+                  }}
                   disabled={batchResults.length > 0}
                 />
               }
```

sharikak54:
When entering scorecards, it is often necessary to refresh the page (e.g. to get new data without accidentally overwriting someone else's work). Currently this causes batch mode to revert to the off state (assuming you had no in-progress batch, which is usually the case when refreshing). At a venue with poor internet (and/or high data needs), this causes extra frustration when accidentally entering the first scorecard after refreshing.

Fixes https://github.com/thewca/wca-live/issues/273.

### DIFF
--- a/client/src/components/admin/AdminRound/AdminRoundContent.jsx
+++ b/client/src/components/admin/AdminRound/AdminRoundContent.jsx
@@ -41,12 +41,15 @@ const ENTER_RESULTS = gql`
 `;
 
 const IS_BATCH_MODE_KEY = `wca-live:is-batch-mode`;
+
 /**
  * Persist whether batch mode is toggled on, in case people navigate
  * or refresh the page
+ * @returns {boolean}
  */
 function getStoreIsBatchMode() {
-  return localStorage.getItem(IS_BATCH_MODE_KEY);
+  const json = localStorage.getItem(IS_BATCH_MODE_KEY);
+  return json ? JSON.parse(json) : false;
 }
 
 function setStoreIsBatchMode(isBatchMode) {
@@ -57,7 +60,6 @@ function setStoreIsBatchMode(isBatchMode) {
  * Persist batch results in local storage, in case people navigate
  * or refresh the page
  */
-
 function getStoreBatchResults(roundId) {
   const json = localStorage.getItem(`wca-live:batch-results:${roundId}`);
   return json ? JSON.parse(json) : [];
@@ -87,7 +89,7 @@ function AdminRoundContent({ round, competitionId, officialWorldRecords }) {
     getStoreBatchResults(round.id),
   );
   const [isBatchMode, setIsBatchMode] = useState(
-    batchResults.length > 0 || getStoreIsBatchMode(),
+    () => batchResults.length > 0 || getStoreIsBatchMode(),
   );
   const formContainerRef = useRef(null);
 
--- a/client/src/components/admin/AdminRound/AdminRoundContent.jsx
+++ b/client/src/components/admin/AdminRound/AdminRoundContent.jsx
@@ -53,7 +53,7 @@ function getStoreIsBatchMode() {
 }
 
 function setStoreIsBatchMode(isBatchMode) {
-  localStorage.setItem(IS_BATCH_MODE_KEY, isBatchMode);
+  localStorage.setItem(IS_BATCH_MODE_KEY, JSON.stringify(isBatchMode));
 }
 
 /**
--- a/client/src/components/admin/AdminRound/AdminRoundContent.jsx
+++ b/client/src/components/admin/AdminRound/AdminRoundContent.jsx
@@ -50,7 +50,7 @@ function getStoreIsBatchMode() {
 }
 
 function setStoreIsBatchMode(isBatchMode) {
-  localStorage.setItem(IS_BATCH_MODE_KEY, isBatchMode)
+  localStorage.setItem(IS_BATCH_MODE_KEY, isBatchMode);
 }
 
 /**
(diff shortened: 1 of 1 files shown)


client/src/components/admin/AdminRound/AdminRoundContent.jsx
```diff
@@ -71,7 +86,9 @@ function AdminRoundContent({ round, competitionId, officialWorldRecords }) {
   const [batchResults, setBatchResults] = useState(() =>
     getStoreBatchResults(round.id),
   );
-  const [isBatchMode, setIsBatchMode] = useState(batchResults.length > 0);
+  const [isBatchMode, setIsBatchMode] = useState(
+    batchResults.length > 0 || getStoreIsBatchMode(),
```
jonatanklosko: We can pass an initialiser function to `useState`. The advantage is that it runs only once to get the default value, and in this case it's preferable, since we don't want to read from local storage on every render.

```suggestion
    () => batchResults.length > 0 || getStoreIsBatchMode(),
```

client/src/components/admin/AdminRound/AdminRoundContent.jsx
```diff
@@ -40,8 +40,23 @@ const ENTER_RESULTS = gql`
...
+/**
+ * Persist whether batch mode is toggled on, in case people navigate
+ * or refresh the page
+ */
+function getStoreIsBatchMode() {
+  return localStorage.getItem(IS_BATCH_MODE_KEY);
+}
+
+function setStoreIsBatchMode(isBatchMode) {
+  localStorage.setItem(IS_BATCH_MODE_KEY, isBatchMode);
```
jonatanklosko: When storing values in local storage, they are converted to string, so if we pass boolean it will actually be stored as `"true"` or `"false"`. Then, `getItem` will also return a string, so it will be treated as a truthy value, which is not what we want.

To fix this, you can use JSON, as in the function below :)

jonatanklosko (commented):
Fantastic, thanks for the PR! I dropped two comments :)

sharikak54:
> We can pass an initialiser function to `useState`. The advantage is that it runs only once to get the default value, and in this case it's preferable, since we don't want to read from local storage on every render.

TIL!  

Thanks for your comments!  I pushed a commit addressing both :)

client/src/components/admin/AdminRound/AdminRoundContent.jsx
```diff
@@ -40,9 +40,26 @@ const ENTER_RESULTS = gql`
...
+ * or refresh the page
+ * @returns {boolean}
+ */
+function getStoreIsBatchMode() {
+  const json = localStorage.getItem(IS_BATCH_MODE_KEY);
+  return json ? JSON.parse(json) : false;
+}
+
+function setStoreIsBatchMode(isBatchMode) {
+  localStorage.setItem(IS_BATCH_MODE_KEY, isBatchMode);
```
jonatanklosko: Let's add explicit conversion to JSON. The default conversion to string is equivalent for booleans, but this makes the intent clear:

```suggestion
  localStorage.setItem(IS_BATCH_MODE_KEY, JSON.stringify(isBatchMode));
```

jonatanklosko (approved):
Looks great, thank you :)

PR #281, opened 2025-07-08
Batch mode highlight

sharikak54:
Make it easier to tell at a glance whether you're currently entering scores in batch mode.  This can be a problem on smaller screen sizes, or when a user has increased their font size: the Batch Mode checkbox is not in view while the input boxes are focused.

I am ambivalent on the color, just wanted something that's bright enough colorblind/vision-impaired people can still notice it but not so bright that it's annoying.

### References

Small screen dark mode: 
Screenshot 2025-07-08 at 12 05 47 PM

Larger screen light mode:
Screenshot 2025-07-08 at 12 01 38 PM

Refs https://github.com/thewca/wca-live/issues/273.

### DIFF
--- a/client/src/components/admin/AdminRound/AdminRoundContent.jsx
+++ b/client/src/components/admin/AdminRound/AdminRoundContent.jsx
@@ -169,10 +169,18 @@ function AdminRoundContent({ round, competitionId, officialWorldRecords }) {
 
   return (
     <>
-      <Grid container direction="row" spacing={2}>
-        <Grid 
-          sx={isBatchMode ? {backgroundColor: "rgba(238, 210, 2, 0.2)"} : {}}
-          item xs={12} md={3} ref={formContainerRef}>
+      <Grid sx={{ marginLeft: "-8px" }} container direction="row" spacing={2}>
+        <Grid
+          sx={{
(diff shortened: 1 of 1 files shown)


jonatanklosko:
Hey! Actually, I think I prefer the original solution that you proposed, which is to store the choice in local storage. I expect that once someone decides to use batch mode, they probably want to use it all around during the competition. What do you think?

jonatanklosko:
Oh, I see you opened #280, nevermind! Personally, I think that should be sufficient, and I find the background a bit arbitrary as an indicator :D

sharikak54:
I think both could be useful!  E.g. at Worlds this weekend, live data entry usually uses non-batch mode.  So if you need to do non-live after (not common, but we had religious exceptions who were solving at odd times), it would be a nicer user experience for your batch mode choice to be obvious.

sharikak54:
I am also totally happy to close this one in favor of #280, which should solve the majority of issues.  If necessary this can always be reopened later 🙂

jonatanklosko:
Yeah, I would revisit later if someone complains. I would rather do a different indicator, but I don't have any obvious ideas right now. We could move it to the top, but I don't think that makes sense either.

Issue #274, opened 2025-07-07
Give warning for duplicate results


sarahstrong314:
This is a feature request. It would be nice if there were a pop-up warning that the results that you entered are an exact match with existing results in the same round. Duplicate results typically happen when the person who entered the results for the wrong person by typoing their ID or selecting the wrong name by accident and overwrites existing results. Duplicate results are often overlooked when the duplicate results overwrite already-double-checked results.

Exceptions would include FMC results and DNF results since ties are common under these circumstances.

PR #282, opened 2025-07-08
Give warning for duplicate results

sharikak54:
I saw this issue https://github.com/thewca/wca-live/issues/274 and thought I'd give it a try.

> It would be nice if there were a pop-up warning that the results that you entered are an exact match with existing results in the same round. Duplicate results typically happen when the person who entered the results for the wrong person by typoing their ID or selecting the wrong name by accident and overwrites existing results. Duplicate results are often overlooked when the duplicate results overwrite already-double-checked results.
>
> Exceptions would include FMC results and DNF results since ties are common under these circumstances.

Screenshot 2025-07-08 at 2 04 43 PM

### ~~Breaking change~~

~~In order to check for duplicate results in the current stored batch, I changed the format of the stored batch results in localStorage to include the `person` attribute.  I figured that made more sense than fetching it each time, since theoretically this should affect few people (only data batches which are in-progress before deployment and need to be refreshed after); but maybe this is a rare enough case that it's better not to introduce a breaking change.  What do you think?~~

Removed after discussion, in favor of creating a `combinedResults` array

### DIFF
--- a/client/src/components/admin/AdminRound/AdminRoundContent.jsx
+++ b/client/src/components/admin/AdminRound/AdminRoundContent.jsx
@@ -59,6 +59,23 @@ function setStoredBatchResults(roundId, results) {
   }
 }
 
+/**
+ * Combine results and batchResults to produce a temporary results-like array
+ * that has all batchResults changes applied.  This allows us to display a
+ * warning if a result is being entered which matches a batched result exactly.
+ */
+function combineResultsAndBatchResults(results, batchResults) {
+  const combinedResults = [...results];
+  batchResults.forEach((batchedResult) => {
+    const i = combinedResults.findIndex((res) => res.id === batchedResult.id);
+    combinedResults[i] = {
+      ...combinedResults[i],
+      attempts: batchedResult.attempts,
+    };
+  });
+  return combinedResults;
+}
+
 function AdminRoundContent({ round, competitionId, officialWorldRecords }) {
   const confirm = useConfirm();
   const { enqueueSnackbar, closeSnackbar } = useSnackbar();
@@ -72,6 +89,9 @@ function AdminRoundContent({ round, competitionId, officialWorldRecords }) {
     getStoreBatchResults(round.id),
   );
   const [isBatchMode, setIsBatchMode] = useState(batchResults.length > 0);
+  const [combinedResults, setCombinedResults] = useState(() =>
+    combineResultsAndBatchResults(round.results, batchResults),
+  );
   const formContainerRef = useRef(null);
 
   const [enterResults, { loading }] = useMutation(ENTER_RESULTS, {
@@ -86,11 +106,11 @@ function AdminRoundContent({ round, competitionId, officialWorldRecords }) {
     onError: apolloErrorHandler,
   });
 
-  function handleResultAttemptsSubmit(attempts, person) {
+  function handleResultAttemptsSubmit(attempts) {
     if (isBatchMode) {
       setBatchResults([
         ...batchResults.filter((result) => result.id !== editedResult.id),
-        { id: editedResult.id, attempts, person, enteredAt: nowISOString() },
+        { id: editedResult.id, attempts, enteredAt: nowISOString() },
       ]);
       setEditedResult(null);
     } else {
@@ -165,6 +185,9 @@ function AdminRoundContent({ round, competitionId, officialWorldRecords }) {
 
   useEffect(() => {
     setStoredBatchResults(round.id, batchResults);
+    setCombinedResults(
+      combineResultsAndBatchResults(round.results, batchResults),
+    );
   }, [round.id, batchResults]);
 
   return (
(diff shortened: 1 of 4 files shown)


client/src/lib/attempt-result.js
```diff
@@ -482,6 +483,23 @@ export function attemptResultsWarning(
         };
       }
     }
+
+    // Check whether this result is a duplicate of existing results in the same round.
+    // Excludes FMC and all-DNF results since ties are common.
+    // TODO: Does NOT check the in-progress batch, if any.
+    if (["333fm"].indexOf(eventId) === -1) {
```
jonatanklosko: Above we have `completeAttempts` with filtered-out DNFs, so we can ignore those results upfront:

```suggestion
    if (eventId !== "333fm" && completeAttempts.length > 0) {
```

Then we no longer need the DNF logic in findAllMatchingResults.

client/src/lib/attempt-result.js
```diff
@@ -482,6 +483,23 @@ export function attemptResultsWarning(
         };
       }
     }
+
+    // Check whether this result is a duplicate of existing results in the same round.
+    // Excludes FMC and all-DNF results since ties are common.
+    // TODO: Does NOT check the in-progress batch, if any.
```
```suggestion
```

client/src/components/admin/AdminRound/AdminRoundContent.jsx
```diff
@@ -86,11 +86,11 @@ function AdminRoundContent({ round, competitionId, officialWorldRecords }) {
     onError: apolloErrorHandler,
   });
 
-  function handleResultAttemptsSubmit(attempts) {
+  function handleResultAttemptsSubmit(attempts, person) {
     if (isBatchMode) {
       setBatchResults([
         ...batchResults.filter((result) => result.id !== editedResult.id),
-        { id: editedResult.id, attempts, enteredAt: nowISOString() },
+        { id: editedResult.id, attempts, person, enteredAt: nowISOString() },
```
jonatanklosko: I don't think we should be storing the whole `person` object in local storage.

What we could do is, right before calling `attemptResultsWarning`, we can build a copy of `results` with changes applied. We can map over `results` and look for a batch result entry with a matching id, if the is one, we take its attempts.

One issue with checking against batch results is that once we show the warning, it may be confusing, because there will be no visible results for the conflicting person (they are in the batch). But I guess it's better than not showing the warning.

jonatanklosko (commented):
Hey, thanks for looking into this, I dropped some comments :)

client/src/components/admin/AdminRound/AdminRoundContent.jsx
sharikak54: Sounds good!  I decided to create a new `combinedResults` array in the `AdminRoundContent` component so that it can be recalculated only when it needs to change.  Let me know if that looks good to you; I'm not a React expert by any means!

I also realized that this warning would pop up if you pulled up a result and tried to enter it without changing anything.  I saw a couple people using this functionality to refresh `results` data without needing to refresh the entire page.  In the interest of not breaking current users' workflows, [I opted to filter out the current result from the duplicate check](https://github.com/thewca/wca-live/pull/282/commits/4177f552794aff566770a41432448ba41454ac7f).

client/src/components/admin/ResultAttemptsForm/ResultAttemptsForm.jsx
```diff
@@ -91,15 +92,23 @@ function ResultAttemptsForm({
...
     });
   }
 
   function confirmSubmission() {
+    // We don't want to show a duplicate warning if the user is submitting
+    // already-entered results for the correct competitor.  This is often used
+    // as a quick way to refresh data without refreshing the whole page.
+    const resultsToCheck = combinedResults.filter(
+      (res) => res.id !== result.id,
+    );
```
jonatanklosko, line 105: I would compute `combinedResults` together with the filter here.

Computing them upfront, as in the current version, doesn't have any benefit, because it would be recomputed after every submission anyway.

client/src/components/admin/AdminRound/AdminRoundContent.jsx
```diff
@@ -76,6 +76,23 @@ function setStoredBatchResults(roundId, results) {
...
+function combineResultsAndBatchResults(results, batchResults) {
+  const combinedResults = [...results];
+  batchResults.forEach((batchedResult) => {
+    const i = combinedResults.findIndex((res) => res.id === batchedResult.id);
+    combinedResults[i] = {
+      ...combinedResults[i],
+      attempts: batchedResult.attempts,
+    };
+  });
+  return combinedResults;
```
jonatanklosko, line 93: Btw. you can also write it like this:

```suggestion
  return results.map((result) => {
    const batchResult = batchResults.find((batchResult) => batchResult.id === result.id);
    if (batchResult) {
      return { ...result, attempts: batchResult.attempts };
    } else {
      return result;
    }
  );
```

I consider it slightly more direct in terms of readability, though either version is fine :)

PR #283, opened 2025-07-10
Add entered results count to staff table

maxidragon:
I'm not sure if this is the best way to do it in Elixir, happy to hear some suggestions! 

image


lib/wca_live/scoretaking.ex
```diff
@@ -641,4 +642,21 @@ defmodule WcaLive.Scoretaking do
...
+  the corresponding competition.
+  """
+  @spec entered_results_count(%StaffMember{}) :: integer()
+  def entered_results_count(%StaffMember{} = staff_member) do
+    from(result in Result,
+      join: round in assoc(result, :round),
+      join: competition_event in assoc(round, :competition_event),
+      join: competition in assoc(competition_event, :competition),
+      where:
+        competition.id == ^staff_member.competition_id and
```
jonatanklosko, line 657: I added competition to the query, because otherwise we would count all results for the given user across all competitions :)

jonatanklosko:
I moved the query to a relevant module, which we then invoke from the resolver. I also added a test to make sure it correctly counts only for the given competition.

Thanks for the PR :)

PR #289, opened 2025-10-22
Allow to check who entered the scorecard

maxidragon:
This is only visible for delegates/organizers.
image

### DIFF
--- a/client/src/components/admin/RoundDoubleCheck/RoundDoubleCheck.jsx
+++ b/client/src/components/admin/RoundDoubleCheck/RoundDoubleCheck.jsx
@@ -20,11 +20,6 @@ const ROUND_QUERY = gql`
       name
       competitionEvent {
         id
-        competition {
-          access {
-            canManage
-          }
-        }
         event {
           id
           name
@@ -208,7 +203,6 @@ function RoundDoubleCheck() {
             disabled={enterLoading}
             onSubmit={handleResultAttemptsSubmit}
             officialWorldRecords={officialWorldRecords}
-            showEnteredBy={round.competitionEvent.competition.access.canManage}
           />
         </Grid>
         <Grid item md sx={{ textAlign: "center" }}>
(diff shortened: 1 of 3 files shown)


maxidragon:
@jonatanklosko kindly ping ;)

client/src/components/admin/RoundDoubleCheck/RoundDoubleCheck.jsx
```diff
@@ -200,6 +208,7 @@ function RoundDoubleCheck() {
             disabled={enterLoading}
             onSubmit={handleResultAttemptsSubmit}
             officialWorldRecords={officialWorldRecords}
+            showEnteredBy={round.competitionEvent.competition.access.canManage}
```
jonatanklosko: Is there a reason not to show it for everyone who can access double check?

lib/wca_live_web/schema/scoretaking_types.ex
```diff
@@ -130,6 +130,10 @@ defmodule WcaLiveWeb.Schema.ScoretakingTypes do
     end
 
     field :entered_at, :datetime
+
+    field :entered_by, :user do
+      resolve dataloader(:db)
+    end
```
jonatanklosko, line 136: Just a note that this will be publicly readable to anyone querying the API, but I wouldn't necessarily consider this a sensitive information, wdyt?

We could make it only work if the querying user can manage the competition, but authorizing specific fields based on user ownership of parent resource (here: the competition) is quite annoying with GraphQL (opposed to, for example, checking global user roles). The basic version of doing `can_access_this?` would result in N+1 queries, so we would need more dataloader magic to make it efficient.

jonatanklosko:
@maxidragon sorry for the late reply, I was off last week :D

lib/wca_live_web/schema/scoretaking_types.ex
maxidragon: (answering both comments here)

I weren't sure whether this is sensitive information, I restricted it to the organizers/delegates just in case and to be consistent since they are the only ones who can access the page which contains information about the total number of entered scorecards.

After thinking about it more, I don't consider this a sensitive information either. I'll modify the code to show this field to all score-takers

jonatanklosko (approved):

PR #291, opened 2025-11-04
Display entered_by only in the double check view

maxidragon:
After merging #289 I checked one of my comps on prod and noticed that the admin round page goes blank, because `enteredBy` is null. It happens, because we don't fetch enteredBy in this view - as it's not needed. I added the check for its presence, which fixes the issue.


jonatanklosko:
Thanks for the quick fix.

Issue #133, opened 2023-01-02
Feature request: scorecard checking with multiple scorekeepers


timreyn:
For major competitions with multiple scorekeepers, it would be great if checking mode sorted by the person who entered the result. So if they entered results in this order
Person A enters scorecard 1
Person B enters scorecard 2
Person A enters scorecard 3
Person A enters scorecard 4
Person B enters scorecard 5

then scorecard checking mode would show them in order 1, 3, 4, 2, 5

The motivation is that it would be easier to keep the scorecards in separate piles per person who entered them, rather than try to keep them sorted in the order they were entered.

For CubingUSA Nationals 2023, we're planning to have 10 scorekeepers, and this would make scorecard checking much easier.

TCiras:
After doing data at both CubingUSA Nationals 2023 and Cubing USA New England Championship 2023, having the checking data in separate lists for different scorekeepers would save time double checking data when getting ready for the nest round of an event to happen.  Sifting through and checking based on ID number rather than looking through the stack in the order it is in takes far more time and slows down the time frame of being ready for that next round.  

This would also make it easier when working with multiple stages, as we could have one scorekeeper per stage which would allow faster location of missing or mistaken data when that issue arises.

kydadinoWCA:
I'm bumping this thread.

Expanding onto this idea, with the addition of 'Batch Mode' a more useful implementation would be to have them ordered by the Batch submission time, followed by the scorecard's entered time. This would allow for multiple scoretakers entering simultaneously, then:

Scenario 1:  The scoretakers can submit their batch one-by-one and move the scorecards into a central pile. They can then pass this off to be stored and/or double checked.

Scenario 2: The scoretakers can submit their batch, and press 'Double Check' right away. This will make the double checking page be at the start of their batch, where they can check until their batch is finished, at which point they are done.

Scenario 3: The scoretakers can submit their batch, and put their batch to the side. This allows them to be passed off to various double checkers, and all they have to do is skip people until they find the start of their batch, then double check as usual. This is similar in practise to Scenario 1.

I'm not sure how easy/difficult this would be to implement, but it would allow for _much more efficient_ double checking at large competitions, including major championships.

kr-matthews:
What's wrong with keeping scorecards in separate piles per person who entered them, right now? For checking, you can just skip through the entries you don't have, and check all the cards in a single pile at a time.

[...]

PR #295, opened 2026-01-03
Allow user to filter by scoretaker when double checking

nick-ng:
In issue https://github.com/thewca/wca-live/issues/133, a suggested solution was:
> But I think a better approach could be having a select where you pick the scoretaker and it shows the results they entered (and the select defaults to "All").

The scoretaker filter only appears when there is more than one scoretaker.

image
image
image

### DIFF
--- a/client/src/components/admin/RoundDoubleCheck/RoundDoubleCheck.jsx
+++ b/client/src/components/admin/RoundDoubleCheck/RoundDoubleCheck.jsx
@@ -217,33 +217,35 @@ function RoundDoubleCheck() {
           </IconButton>
         </Grid>
         <Grid item md={3}>
-          <FormControl fullWidth sx={{ mb: 2 }}>
-            <InputLabel id="scoretaker-filter-label">
-              Filter Scoretaker
-            </InputLabel>
-            <Select
-              labelId="scoretaker-filter-label"
-              id="scoretaker-filter"
-              value={scoretakerFilter}
-              label="Filter Scoretaker"
-              onChange={(event) => {
-                updateScoretakerFilter(event.target.value);
-                updateResultIndex(0);
-                setTimeout(() => {
-                  rightButtonRef.current.focus();
-                }, 0);
-              }}
-            >
-              <MenuItem value={""}>
-                <em>All</em>
-              </MenuItem>
-              {scoreTakerNames.map((name) => (
-                <MenuItem key={name} value={name}>
-                  {name}
+          {scoreTakerNames.length > 1 && (
+            <FormControl fullWidth sx={{ mb: 2 }}>
+              <InputLabel id="scoretaker-filter-label">
+                Filter Scoretaker
+              </InputLabel>
+              <Select
+                labelId="scoretaker-filter-label"
+                id="scoretaker-filter"
+                value={scoretakerFilter}
+                label="Filter Scoretaker"
+                onChange={(event) => {
+                  updateScoretakerFilter(event.target.value);
+                  updateResultIndex(0);
+                  setTimeout(() => {
+                    // rightButtonRef.current.focus();
+                  }, 0);
+                }}
+              >
+                <MenuItem value={""}>
+                  <em>All</em>
                 </MenuItem>
-              ))}
-            </Select>
-          </FormControl>
+                {scoreTakerNames.map((name) => (
+                  <MenuItem key={name} value={name}>
+                    {name}
+                  </MenuItem>
+                ))}
+              </Select>
+            </FormControl>
+          )}
           <ResultAttemptsForm
             result={results[resultIndex]}
             results={round.results}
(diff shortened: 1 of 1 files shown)


client/src/components/admin/RoundDoubleCheck/RoundDoubleCheck.jsx
```diff
@@ -192,6 +215,33 @@ function RoundDoubleCheck() {
...
+              labelId="scoretaker-filter-label"
+              id="scoretaker-filter"
+              value={scoretakerFilter}
+              label="Filter Scoretaker"
+              onChange={(event) => {
+                updateScoretakerFilter(event.target.value);
+                updateResultIndex(0);
+                setTimeout(() => {
+                  rightButtonRef.current.focus();
+                }, 0);
```
nick-ng: The `<Select>` becomes focused after choosing an item. When you choose "All", it looks like this
Image

By focusing something else, the `<Select>` goes back to showing the "Filter Scoretaker" text in the middle of the `<Select>`.

client/src/components/admin/RoundDoubleCheck/RoundDoubleCheck.jsx
```diff
@@ -103,6 +113,12 @@ function RoundDoubleCheck() {
     variables: { id: roundId },
   });
 
+  const filteredResults =
+    data?.round.results.filter((result) => {
+      if (!scoretakerFilter || scoretakerFilter === "0") return true;
+      return result.enteredBy?.name === scoretakerFilter;
+    }) || [];
```
jonatanklosko: If this condition is true then we keep the list as is, in which case we don't need to do filter at all. Let's do something like:

```js
const filteredResults = scoretakerFilter : data.round.results.filter(...) : data.round.results;
```

client/src/components/admin/RoundDoubleCheck/RoundDoubleCheck.jsx
```diff
@@ -103,6 +113,12 @@ function RoundDoubleCheck() {
     variables: { id: roundId },
   });
 
+  const filteredResults =
+    data?.round.results.filter((result) => {
+      if (!scoretakerFilter || scoretakerFilter === "0") return true;
```
jonatanklosko: It should never be "0"?

client/src/components/admin/RoundDoubleCheck/RoundDoubleCheck.jsx
```diff
@@ -192,6 +217,35 @@ function RoundDoubleCheck() {
...
+                labelId="scoretaker-filter-label"
+                id="scoretaker-filter"
+                value={scoretakerFilter}
+                label="Filter Scoretaker"
+                onChange={(event) => {
+                  updateScoretakerFilter(event.target.value);
+                  updateResultIndex(0);
+                  setTimeout(() => {
+                    // rightButtonRef.current.focus();
+                  }, 0);
```
jonatanklosko:
```suggestion
```

client/src/components/admin/RoundDoubleCheck/RoundDoubleCheck.jsx
```diff
@@ -130,8 +146,17 @@ function RoundDoubleCheck() {
...
   const { round, officialWorldRecords } = data;
 
+  const scoreTakerNames = [
+    ...round.results.reduce((acc, result) => {
+      if (result.enteredBy?.name) {
+        acc.add(result.enteredBy.name);
+      }
+      return acc;
+    }, new Set()),
+  ].sort((a, b) => a.localeCompare(b));
```
jonatanklosko: It would be better to use ids as select value, which are by definition unique. To do that we can build a unique list of scoretakers by building `{[id]: [scoretaker]}` object and then doing `Object.values` + sort.

jonatanklosko (commented):
Thanks for the PR, I dropped some code comments, functionality-wise looks good to me :)

client/src/components/admin/RoundDoubleCheck/RoundDoubleCheck.jsx
nick-ng: Oops. I was trying different things for when the user hadn't selected anything.

client/src/components/admin/RoundDoubleCheck/RoundDoubleCheck.jsx
nick-ng: Good idea

client/src/components/admin/RoundDoubleCheck/RoundDoubleCheck.jsx
nick-ng: Good idea

client/src/components/admin/RoundDoubleCheck/RoundDoubleCheck.jsx
```diff
@@ -130,8 +148,17 @@ function RoundDoubleCheck() {
...
   const { round, officialWorldRecords } = data;
 
+  const scoretakers = Object.values(
+    round.results.reduce((acc, result) => {
+      if (result.enteredBy) {
+        acc[result.enteredBy.id] = result.enteredBy;
+      }
+      return acc;
+    }, {}),
+  ).sort((a, b) => a.name.localeCompare(b.name));
```
nick-ng, line 158: I considered adding the scoretaker's WCA ID to the `<Select>` so you had a way of differentiating scoretakers with identical names but it seemed unnecessary.

It's unlikely for a competition to have so many scoretakers who also have identical names like the example data.

jonatanklosko (approved):
Thanks :)

PR #307, opened 2026-08-02
Add quit competitors audit log and history view

Victus0660:
## Description
This PR resolves a pain point during scoretaking double-checking:
1. **Double-checking quit competitors**: Allows scoretakers to easily confirm competitors who were quit/removed from a round via a dedicated history list, rather than checking for absence.
2. **Audit & Accountability**: Provides Delegates with visibility into who performed each quit action, when it occurred, and whether a replacement competitor was added.

## Changes
- Added `round_removals` database migration and `RoundRemoval` Ecto schema.
- Updated `remove_person_from_round` and `remove_no_shows_from_round` to log audit entries with `current_user`.
- Exposed `removals` field on `:round` GraphQL schema.
- Added `QuitHistoryDialog` component and integrated "Quit history" button into `AdminRoundToolbar`.

### DIFF
--- a/client/src/components/admin/AdminRound/AdminRoundToolbar.jsx
+++ b/client/src/components/admin/AdminRound/AdminRoundToolbar.jsx
@@ -5,8 +5,10 @@ import PersonAddIcon from "@mui/icons-material/PersonAdd";
 import CheckIcon from "@mui/icons-material/Check";
 import DeleteSweepIcon from "@mui/icons-material/DeleteSweep";
 import PrintIcon from "@mui/icons-material/Print";
+import HistoryIcon from "@mui/icons-material/History";
 import AddCompetitorDialog from "./AddCompetitorDialog";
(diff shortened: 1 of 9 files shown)


jonatanklosko:
Hey @Victus0660, thanks for the PR! However, there is ongoing work on a live results platform integrated directly into the main WCA website (Integrated Live Results), and I believe the plan is to fully transition to that system in the upcoming months. For this reason, I'd rather avoid adding new features of this scope to WCA Live at this point.

You may have a look at https://github.com/thewca/worldcubeassociation.org. It may be a suitable to discuss the feature there, or it may even already be in place.

Victus0660:
@jonatanklosko 
Thanks for the explanation and feedback! I completely understand the reasoning with Integrated Live Results on the way. I'll check out the main repository and see how I can contribute there instead. Cheers!

PR #309, opened 2026-08-23
Preserve selected schedule day when navigating away

maxidragon:
This PR fixes probably one of the most annoying issues in WCA Live. Currently, when a user opens a round that isn't on the first day of the competition and then goes back, they are taken back to the first day of the competition. After these changes, the selected day is preserved until the user returns to the homepage.


jonatanklosko:
I agree that it's an issue, but I feel like the current solution has too many parts.

What I would suggest is that the schedule stores `{competitionId,date}` json inside session storage, the on mount it reads from storage and if the competition id is the same it restores the date, otherwise it can store the new competition, plus whenever the date tab changes. This should keep it all in a single file.

client/src/components/Schedule/Schedule.jsx
```diff
@@ -33,7 +48,26 @@ function Schedule({ venues, competitionEvents, competitionId }) {
     sortedActivities.map((activity) => toLocalDateString(activity.startTime)),
   );
 
-  const [selectedDate, setSelectedDate] = useState(closestDateString(dates));
+  // We keep the selected date in the URL, so that it's preserved
+  // when the user navigates to a round and goes back.
+  const [searchParams, setSearchParams] = useSearchParams();
```
jonatanklosko: With the storage we don't need url params anymore.

client/src/components/Schedule/Schedule.jsx
```diff
@@ -10,6 +10,21 @@ import {
 import { eventRoundForActivityCode } from "../../lib/competition";
 import { parseActivityCode } from "../../lib/activity-code";
 
+const SCHEDULE_DATE_KEY = "scheduleDate";
+
+function getStoredScheduleDate(competitionId, dates) {
```
jonatanklosko: Let's move this below the component.

client/src/components/Schedule/Schedule.jsx
```diff
@@ -10,6 +10,21 @@ import {
...
+  try {
+    const stored = JSON.parse(sessionStorage.getItem(SCHEDULE_DATE_KEY));
+    if (
+      stored?.competitionId === competitionId &&
+      dates.includes(stored?.date)
+    ) {
+      return stored.date;
+    }
+  } catch {}
```
jonatanklosko: To fix the lint we can `return null` in the catch. and above, instead of at the end.

jonatanklosko (approved):

